PR #463, opened 2014-06-16
Solving 'unlock_at' edition bug for modules

Kulgar:
Hi :)

A new bug to begin this wonderful week! ^^
This concerns the "unlock_at" date for modules and translations. I'm offering a pull request but I'm not sure it is the right way to solve this bug (but it does work). 

## :1234: Test plan

* Create a course, add a module,
* Change your language to whatever... (French for instance, as I use French most of the time),
* Edit the module and check the "unlock this module" checkbox, 
* Add an unlock at date (in the future) and save
* Edit again the module, you should see something like "this isn't a date" (or in French: Merci de choisir une date)
* If you edit something else and save (maybe after refreshing the page), the 'unlock at' date will disappear and thus the module won't unlock at the first inserted date! 

This is due to date translations & date formatting. French format (and I think other languages too) is not recognized when the date is saved, but as you fill in the form using the footer of the module, you fill it with the translation of the date and thus it creates the bug. 

## My fix

I added a new hidden span in the views where I insert the unformatted and untranslated date. Then I get it back from "context_modules.js" when we edit the module and fill in the form using that value instead of the one stored in the "unlock_at" span. 

I also get the untranslated value when we are comparing the date with the actual date for the show/hide "unlock details" information of the footer, because, again, the French translated date (and other translated dates) is not recognized by the Date.parse javascript function (it returns "null" so the test always fails). 

Let me know if I can improve my solution. :)

Kulgar.

### DIFF
--- a/app/views/context_modules/_context_module.html.erb
+++ b/app/views/context_modules/_context_module.html.erb
@@ -97,6 +97,7 @@
       <span class="progression_state" style="<%= hidden if editable || !context_module || ((context_module.prerequisites || []).empty? && (context_module.completion_requirements || []).empty?) %>"></span>
       <div class="unlock_details" style="<%= hidden unless context_module && context_module.to_be_unlocked %>">
         <%= t 'module_will_unlock_at', 'Will unlock *%{unlock_date}*', :wrapper => "<span class=\"unlock_at\" #{context_sensitive_datetime_title(context_module.try_rescue(:unlock_at), @context)}>\\1</span>", :unlock_date => (datetime_string(context_module.try_rescue(:unlock_at)) || nbsp) %>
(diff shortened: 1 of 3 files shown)


simonista:
Hey @Kulgar, is this still something you're interested in finishing up? Or should we just close it for now?

Kulgar:
Hey @simonista - in fact I thought you already solved that in another update... 
So I did not know it was still something I should continue. But sure, I do want to finish this up if you're still interested.

simonista:
It may in fact be something we already solved, I was just looking through old pull requests.  Let me look and see.

simonista:
Yes, this does indeed appear to be solved now.  Thanks anyway for the pull request!

PR #491, opened 2014-08-06
LTI: add assignment dates as custom parameters

hballweg:
By sending assignment dates if specified, LTI tool providers can use those dates to e. g. change the behaviour of the tool after the due date.

I didn't provide a test plan as is, but I'm happy to provide one if necessary.


simonista:
This work has now already been done and is no longer needed.  Thanks for the suggestion and push though.

Issue #531, opened 2014-10-27
Suggested improvement to Canvas Conference’s Interface: Delete past conference


jfederico:
Deleting a past conference deletes recordings associated

simonista:
Fixed in https://github.com/instructure/canvas-lms/pull/532

Issue #530, opened 2014-10-27
Suggested improvement to Canvas Conference’s Interface: End a session


jfederico:
When a conference is In Progress, add ability to end the conference

![end_conference](https://cloud.githubusercontent.com/assets/578359/4785525/f48f0680-5d7f-11e4-8f3c-d1a1f259f007.PNG)

This allows the teacher the ability to kick everyone out of a session after they have left.

simonista:
Fixed in https://github.com/instructure/canvas-lms/pull/532

Issue #529, opened 2014-10-27
Suggested improvement to Canvas Conference’s Interface: Delete a session


jfederico:
When deleting a session
–Also ends the session on the server
![delete_conference](https://cloud.githubusercontent.com/assets/578359/4785490/9659c71c-5d7f-11e4-9841-90e7b07f3a37.png)

simonista:
Fixed in https://github.com/instructure/canvas-lms/pull/532

PR #532, opened 2014-10-27
Suggested improvements to Canvas Conference’s Interface

jfederico:
Suggested code for issues:

https://github.com/instructure/canvas-lms/issues/529
https://github.com/instructure/canvas-lms/issues/530
https://github.com/instructure/canvas-lms/issues/531
https://github.com/instructure/canvas-lms/issues/534


app/models/big_blue_button_conference.rb
```diff
@@ -81,6 +81,49 @@ def recordings
...
+    super
+  end
+
+  def destroy
+    end_meeting
+    delete_all_recordings
+    super
+  end
+
+  set_policy do
```
jfederico: Wit code duplication, do you mean the close/destroy methods?

Regarding to the set_policy, yeah I know what you mean, I tried not to move anything in the superclass because I think you use that class for other conference systems.
In fact the only change required is to ignore long_running for bigbluebutton because it should be possible to complete any meeting.

    given { |user, session| context.grants_right?(user, session, :manage_content) && active? }
    can :close

jfederico:
There are still a couple of features that need to be added in the client for publish/unpublish and delete recordings, but I decided not to move forward until the changes on the server were accepted.

I can go back to client but I am not sure if this would work better for you guys. I think we should first figure out how to eliminate the duplication in set_policy, but to be hones I don't know how to do that and not to break the code for other conference modules you may have.

jfederico:
Hey Matthew,
Sorry for making you waste your time.

Because the change made to app/coffeescripts/models/Conference.coffee the override of set_policy was totally unnecessary.

The code used now is only the one in the super class

mcwqy9:
Haha no worries jfederico. I had never fully explored the underbelly of our permissions framework and it was a good learning experience.

app/models/big_blue_button_conference.rb
```diff
@@ -117,11 +142,27 @@ def fetch_recordings
 
   def delete_recording(recording_id)
     response = send_request(:deleteRecordings, {
-      :recordID => recording_id,
```
mcwqy9: In Ruby I prefer to leave the terminal comma in a hash or array, because it reduces the super common bug of adding another line to a multiline array or hash and forgetting to add a comma to the previous line, and it makes diffs cleaner because you don't have to add an incidental change to the previous line.

app/models/big_blue_button_conference.rb
```diff
@@ -117,11 +142,27 @@ def fetch_recordings
...
+      :recordID => recording_id
       })
     response[:deleted] if response
   end
 
+  def publish_recording(recording_id)
+    response = send_request(:publishRecordings, {
+      :recordID => recording_id,
```
mcwqy9: This line and the next should be intended to the same level.

app/models/big_blue_button_conference.rb
```diff
@@ -117,11 +142,27 @@ def fetch_recordings
...
+	  :publish => true
+      })
+    response[:published] if response
+  end
+
+  def unpublish_recording(recording_id)
+    response = send_request(:publishRecordings, {
+      :recordID => recording_id,
```
mcwqy9: This line and the next should be intended to the same level.

mcwqy9:
Overall I'm ok with this. Green light to proceed with the front end work.

We'll need some automated unit tests around the new methods on BigBlueButtonConference.

We (Instructure) have a pretty effective hybrid manual-automated QA process we always run changes through, but at this point it would mostly be a regression test of BBB conferences in Canvas. I'd prefer to see the front end written up and tests added (aka feature, code, and test complete) before merging this into master so that we only have to regression test this once. Sound good?

mcwqy9:
and I'll do much better in future at getting back to you more quickly, my apologies.

jfederico:
Hi Matt,
Thanks for the feedback. I certainly learned a couple of things here! Changes applied.

Regarding to the waiting until the changes on the client are made, would you reconsider doing the merge with what we got so far?

The thing is that these are not new features but fixes to old bugs that according to the feedback we (the BigBlueButton developers) have been receiving, are causing real problems to end users.

The change to the client is basically just one and it is to add capabilities for managing recordings (delete, publish and unpublish), so we can consider this as a new feature, that  I think it may take a little longer to complete. 
This because first we need to agree on where are we going to put these options, (buttons or menu) and second, as I am not very familiar yet with how the client works it will take me some more time. 

I actually have a couple of questions I'll better send you by email.

Thanks again

mcwqy9:
Ok, I see what you are saying about getting some bug fixes in. In that case, would you consider removing the new publish_recording & unpublished_recording methods to a different pull request? Then I would be comfortable merging the remaining code.

app/coffeescripts/models/Conference.coffee
```diff
@@ -19,7 +19,7 @@ define [
 
     permissions_data: ->
       has_actions: @get('permissions')['edit'] || @get('permissions')['delete']
-      show_end: @get('permissions')['initiate'] && @get('started_at') && @get('long_running')
+      show_end: @get('permissions')['initiate'] && @get('permissions')['close'] && @get('started_at')
```
JoelHough: This check can just be ```@get('permissions')['close']```. That's what the controller checks on the other side of the api call.

jfederico: Agreed, change applied

jorgevbo:
+1
I'm working in a Rails worker to check BBB running conferences using BBB API and I've encountered problem with Canvas LMS, because it isn't closing web conferences. :(

jfederico:
@JoelHough Done. All seems to be working fine now.

simonista:
Done in 013999f7d9fe21d9b33733cb6dc1af1c01742f14

PR #553, opened 2014-12-01
Warn about setting the password of a SSO user in an SIS import

eriko:
Setting the password on a user when you are using a SSO or LDAP will reset the users session when the SIS import happens.
Per 715541 zendesk ticket.


ccutrer:
merged #554 instead

PR #554, opened 2014-12-01
Warn about the combination of SIS, passwords, and LDAP or an SSO protoco...

eriko:
Ne version of #553


eriko:
While I have not reactivated my enrollment processing code that sets a password I thought this issue had been fixed.  I mean really who wants to create an account with out setting a password.

PR #602, opened 2015-03-09
preventing course[event] arbitrary cast to symbol

geekspeed:
Using a guard function in front of the cast from params to_sym prevents an unsafe cast to symbol. This prevents a nasty DOS condition in which an arbitrary amount of symbols could be created and exhaust all system memory.

### DIFF
--- a/app/controllers/conversations_controller.rb
+++ b/app/controllers/conversations_controller.rb
@@ -942,8 +942,14 @@ class ConversationsController < ApplicationController
   end
 
   def infer_scope
-    filter_mode = (params[:filter_mode].respond_to?(:to_sym) && params[:filter_mode].to_sym) || :or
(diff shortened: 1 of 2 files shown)


geekspeed:
Fixed another potential issue in filter_mode inside conversations_controller.

simonista:
Hey @geekspeed, thanks for the pull request!  A couple quick things:

1. Would you mind filling out a Contributor License Agreement (https://github.com/instructure/canvas-lms/wiki/ica.pdf), and emailing it to me (simon@instructure.com).  We'll need that before we can incorporate this PR.

2. Could you re-create this pull request against our master branch, instead of the stable branch?

Thanks!

geekspeed:
Resubmitted. Thanks.

PR #603, opened 2015-03-09
Fixed a conversion from params to sym

geekspeed:
Re-arranged the conditional logic so we don't accidentally convert arbitrary strings inside params[:filter_mode] to symbols.

### DIFF
--- a/app/controllers/conversations_controller.rb
+++ b/app/controllers/conversations_controller.rb
@@ -942,8 +942,14 @@ class ConversationsController < ApplicationController
   end
 
   def infer_scope
-    filter_mode = (params[:filter_mode].respond_to?(:to_sym) && params[:filter_mode].to_sym) || :or
-    return render_error('filter_mode', 'invalid') if ![:or, :and].include?(filter_mode)
+    # John Stauffacher (john.stauffacher@gmail.com) prevents an unsafe cast to sym 
+    if params[:filter_mode].respond_to?(:to_sym)
+    	if ['or','and'].include? params[:filter_mode]
+    		filter_mode = (params[:filter_mode].respond_to?(:to_sym) && params[:filter_mode].to_sym) 
+    	else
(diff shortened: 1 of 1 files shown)


simonista:
Hey @geekspeed, sorry for the radio silence.  I just pulled these into our internal CI system to run through the automated test suite and get in front of the right people for code review.

Do you think it would be possible to write an automated test for this to prevent it from regressing in the future?

app/controllers/conversations_controller.rb
```diff
@@ -942,8 +942,14 @@ def render_error(attribute, message)
...
-    filter_mode = (params[:filter_mode].respond_to?(:to_sym) && params[:filter_mode].to_sym) || :or
-    return render_error('filter_mode', 'invalid') if ![:or, :and].include?(filter_mode)
+    # John Stauffacher (john.stauffacher@gmail.com) prevents an unsafe cast to sym 
+    if params[:filter_mode].respond_to?(:to_sym)
+    	if ['or','and'].include? params[:filter_mode]
+    		filter_mode = (params[:filter_mode].respond_to?(:to_sym) && params[:filter_mode].to_sym) 
+    	else
+    		return render_error('filter_mode', 'invalid')
+	end
+    end
```
simonista: some initial stylistic feedback:
- we prefer 2 spaces over tabs
- remove trailing whitespace

ccutrer:
While symbol pollution is a currently an attack vector on Ruby 1.9.3, which is the current supported version of Ruby for Canvas, we're very close to changing that to Ruby 2.1, and then Ruby 2.2 soon afterwards. At that point, symbol GC makes this no longer a problem. Given that there are likely hundreds to thousands of these attack vectors in current Canvas, we're rather dedicate the time to getting the Ruby 2.2 upgrade done and fix them all at once, than play whack-a-mole and polluting the code with these minor fixes.

PR #604, opened 2015-03-09
Fixed conversion from params to sym in update method

geekspeed:
Update method cast course[:event] to symbol and allowed arbitrary values for course[event] param to pollute the symbol table. This could lead to a DOS attack.


ccutrer:
You'll need to cherry-pick your commits on top of master, and then make that your pull request. As it is now, you're asking to merge 250+ commits from stable onto master.

geekspeed:
Yikes. Not what i was wanting to do.

PR #605, opened 2015-03-09
Fixed a conversion from params to sym in course_controller update method

geekspeed:
Fixed a nasty situation where we course[event] was getting cast to symbol and could potentially pollute the symbol table and consume all running memory. Symbols are not GCd

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -1966,7 +1966,13 @@ class CoursesController < ApplicationController
       end
 
       if params[:course][:event] && @course.grants_right?(@current_user, session, :change_course_state)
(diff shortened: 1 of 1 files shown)


ccutrer:
While symbol pollution is a currently an attack vector on Ruby 1.9.3, which is the current supported version of Ruby for Canvas, we're very close to changing that to Ruby 2.1, and then Ruby 2.2 soon afterwards. At that point, symbol GC makes this no longer a problem. Given that there are likely hundreds to thousands of these attack vectors in current Canvas, we're rather dedicate the time to getting the Ruby 2.2 upgrade done and fix them all at once, than play whack-a-mole and polluting the code with these minor fixes.

PR #609, opened 2015-03-20
Augment module item completion requirements with 'mark as done' (Take 2)

kjellm:
See pull request #571

PR #571, opened 2015-01-21
Augment module item completion requirements with 'mark as done'.

kjellm:
Hi

I am working on a project for the Norwegian government aiming to
create a MOOC (based on Canvas) with several courses for continuing
education of teachers of mathematics.

The courses are module based and relies heavily on the module item
requirements to enable students to track progress. This works great
for the most part, but there is missing a good mechanism for students
to keep track of progress with regards to articles (wiki pages). The
closest thing is the 'must view' requirement, but this does not cater
to students who needs multiple reading sessions before they feel they
are done.

This requirement can also be used on assignments that are purely for
self studying.

This pull-request is to be regarded as a prototype of this 'mark as
done' functionality. My motives for creating an early pull-request is
to hopefully get your feedback on, primarily: Whether this is
functionality you are interested in merging into the project? And
secondarily: Is my implementation close to how you would like this
feature implemented? I would really appreciate any feedback on how
this functionality should be implemented from your point of view. I am
more than willing to change the code accordingly.

Known issues with the prototype:

  - need to write tests.
  - need to document new API-endpoints.
  - fix ugly controller code and duplication.
  - need to do UI design.
  - unchecking the 'mark as done' checkbox does not work (not done
    implementing API endpoint).

Kjell-Magne

PR #610, opened 2015-03-23
Augment module item completion requirements with 'mark as done' (Take 3)

kjellm:
In module based courses, to make it easier for students to keep
track of their progress. This commit adds support for 'mark as
done' requirements to the module progressions. This to better
cater to students who needs multiple reading sessions before they
feel they are done with a module item (i.e wiki pages).

This requirement can also be used on assignments that are purely
for self studying.

(See #571 and #609 for the pull requests history)


claydiffrient:
@kjellm I'm so sorry for the radio silence on this recently.  Things have been a bit hectic, but I have some feedback for you on this.

It didn't pass our tests when I ran it. Looking at the failures, most seem to be related to the problem found below.

When attempting to go to `/courses/x/wiki` locally I'm getting an error.  In the log the error and relevant stack trace is:
```
NoMethodError (undefined method `has_requirement?' for nil:NilClass):
  /Users/cdiffrient/Projects/canvas/app/views/shared/_mark_as_done.html.erb:1:in `_app_views_shared__mark_as_done_html_erb___2479023243758000851_70291362916940'
```

Hopefully you can figure out what's going on with that and make the necessary changes.

kjellm:
@claydiffrient Thank you for the feedback. I will take a look at this on Monday.

claydiffrient:
@kjellm No problem another thing to note as well is that for i18n the use of keys in translations is deprecated because we auto generate keys now.  So anywhere you are using `t('key', 'phrase')` or similar should be changed to be `t('phrase')`.  One particular spot is here: https://github.com/matematikk-mooc/canvas-lms/commit/5e69148b803be7814485dd7025852fe1e15564e0#diff-9e7aa2a0e40380766f085070db428b70R86

kjellm:
@claydiffrient With regards to new usage of the `t()` function. It seems strange to change this in places where the surrounding code still uses the two parameter version. I think it is better to be consistent within a single file. So only changing where this will not cause inconsistencies (i.e. https://github.com/matematikk-mooc/canvas-lms/commit/5e69148b803be7814485dd7025852fe1e15564e0#diff-428e6504eacecdeb5324667f43df3230R5) Let me know if you want me to change all places regardless of it introducing inconsistency.

kjellm:
@claydiffrient 4164f2f should fix the errors. However, I didn't find any test that could verify this (not an exhaustive search). Could you list the failing tests if there stil are problems? Also let me know if I shall squash the new commits and make a new pull request.

claydiffrient:
@kjellm So there are a few things you can use to test your code before hand.  We recently added rubocop as a linter for our ruby changes.  If you run `script/rlint` that will spit out warnings for whatever commit is sitting at `HEAD`.  Anything other than `INFO` warnings need to be fixed.

Another good thing to do is `bundle exec rake i18n:check` which will validate all the translation stuff.  Which I'm okay with you leaving things you didn't touch with the old version of the `t()` however, the linter/i18n:check will fail anything that isn't updated on new code you put in.

kjellm:
Continued in #620

PR #615, opened 2015-04-06
Say "web browser" instead of "internet browser"

alexhenrie:
The term "Internet browser" is not used professionally because is it not
technically correct. Many programs use the Internet: E-mail clients such
as Thunderbird and Outlook, instant messangers, smartphone apps, game
consoles, and much more. Web browsers are just another kind of program
that uses the Internet; they use it specifically to access the
collection of URI-accessible documents known as the World Wide Web.

### DIFF
--- a/app/views/context_modules/url_show.html.erb
+++ b/app/views/context_modules/url_show.html.erb
@@ -15,17 +15,17 @@
     <div>
       <a id="open_url_button" target="_blank" role="link" class="btn external"
          href="<%= @tag.url %>"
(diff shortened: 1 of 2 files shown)


claydiffrient:
@alexhenrie Thanks for the pull request.  This makes a lot of sense to me.  I'm running it through the appropriate review process on our end, but it should be a quick merge.  Have you already submitted a CLA?

alexhenrie:
I emailed my CLA last night, and I have now added a second commit to this pull request that removes the translation keys.

PR #617, opened 2015-04-09
[Solving Bug] Solving send data truncated bug

Kulgar:
Hi everyone, 

I have discovered a new bug and quite a tricky one. 

How to reproduce it:
* find any action controller that uses the "send_data" method (to export a CSV for instance, like gradebook.csv),
* create data with multibytes characters (in users' names for instance), those data must be included in the ones you want to export, 
* export those data, the file you get should miss some characters at the end of the file, you can check the file you get by comparing it to a CSV that you generate using the ruby console, 

One of my exported files (gradebook.csv) was truncated because of multibytes characters in users' names. 

I did a lot of reverse engineering and I finally discovered that you were somewhat supercharging the send_data method in "config/initializers/action_pack.rb" by defining the "Content-Length" header. 

I did some research and came across this issue: https://rails.lighthouseapp.com/projects/8994/tickets/2661-fix-for-send_data-content-length-header-in-ruby-19

This was related to Ruby On Rails 2.3.4 but it is the exact same issue. Looking at their patch I found that for ruby > 1.9 they are now using "bytesize" method instead of "length" for this header. 

So I'm offering this fix which does exactly that and which solves my truncated data issue. 

Let me know if you need any further information,
Regis

Original PR: #616

### DIFF
--- a/config/initializers/action_pack.rb
+++ b/config/initializers/action_pack.rb
@@ -13,7 +13,7 @@ ActionController::DataStreaming.class_eval do
   alias_method_chain :send_file, :content_length
 
   def send_data_with_content_length(data, options = {})
-    headers.merge!('Content-Length' => data.length.to_s) if data.respond_to?(:length)
+    headers.merge!('Content-Length' => data.bytesize.to_s) if data.respond_to?(:bytesize)
     send_data_without_content_length(data, options)
   end
   alias_method_chain :send_data, :content_length
--- a/spec/controllers/gradebooks_controller_spec.rb
+++ b/spec/controllers/gradebooks_controller_spec.rb
@@ -332,6 +332,18 @@ describe GradebooksController do
           Enrollment.expects(:recompute_final_score).never
           get 'show', :course_id => @course.id, :init => 1, :assignments => 1, :format => 'csv'
         end
+        it "should get all the expected datas even with multibytes characters", :focus => true do
+          @course.assignments.create(:title => "Déjà vu")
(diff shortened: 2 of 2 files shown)


ccutrer:
do you think you could come up with a spec to test this?

Kulgar:
I thought you already had a spec to test send_data export results?
For instance, do you have a spec to test gradebook.csv content?

If so, I may be able to develop a spec, yes.

ccutrer:
Yes we have a spec for that, but obviously it doesn't properly test a multi-byte character.

Kulgar:
I created a new rspec task to test multibytes characters in data downloaded through send_data. 

I did it in the Gradebook controller spec, as the gradebook csv is quite important I think. 

As I couldn't get the real downloaded body (the response.body is not the same as the body we get through our web browser), I kinda simulated the downloaded body by using the "limit" method on a multibyte characters string and passing the Content-Length header as the parameter. 

If you think about a better solution to test the real downloaded body, I will modify my rspec. 

Also, if I modify the code, do you want me to make new commits? Or is there a better Instructure way to give you a cleaner pull request?

Thanks, 
Regis.

ccutrer:
an additional commit is fine. I'll send this through code review with the appropriate team at Instructure (I'm not familiar enough with the spec thing to answer your question myself)

ccutrer:
it has passed code review, and is waiting on QA

Kulgar:
Great! (QA is for? Quality Analysis?)

ccutrer:
Yes, quality assurance. Basically, one more check that it does what it says it does

PR #640, opened 2015-06-16
replace slashes with underscores in gradebook export filename

grahamb:
When using local storage instead of S3, having a course short name with a slash in it will cause an error (No such file or directory @ rb_sysopen) when trying to export the Gradebook as a CSV. attachment_fu is receiving the slash as part of the file path, and erroring out because it can't find the file.

A CLA is on file under my employer, Simon Fraser University (github.com/sfu). Happy to review this in person at Instructurecon :)

Test plan:

  - enable local storage for canvas
  - create a course with a slash in the short name
    (e.g. CHEM 121 / CHEM 122)
  - attempt to export the gradebook
  - ensure the gradebook file was exported and downlaoded correctly


claydiffrient:
@grahamb There's been some concern that this fix, while good, might not be the best place to handle this change.  Perhaps it could me moved into the attachment code so that it would apply to everything not just to gradebook csv exports.

Also I'd love to meet you in person at InstCon, I'll be up there tomorrow (and at Hack Night)

grahamb:
Sure, that makes sense. Let's chat tomorrow night about it. 

— Graham

grahamb:
Will open a new PR with a fix in AttachmentFu instead.

PR #644, opened 2015-06-16
fixed example

ArtskydJ:
Fixed a typo, and changed the string to an array.


claydiffrient:
@ArtskydJ Would you mind fixing up your commit message just a bit?  We want our git history to be pretty clean and explain each change well.  You could simply provide a bit more elaboration and say something like: "fixed documentation example in SIS import api" which would help us identify the change quickly later on if needed.

ArtskydJ:
Not quite sure how to change the commit message, so I made PR #645. It has the same code, but a better message.

PR #645, opened 2015-06-16
Fixed documentation example in SIS import api

ArtskydJ:
- Fixed a typo.
- Changed the array example from a string representation of an array to an array.


claydiffrient:
Uh oh... it looks like you built this one against stable rather than master.

ArtskydJ:
LOL, sorry, apparently I'm bad at this: #646.

claydiffrient:
Don't worry @ArtskydJ It happens to the best of us :)

PR #647, opened 2015-06-19
Update code of conduct to integrate Contributor Covenant 1.1.0

strand:


strand:
@claydiffrient requested I fix this internally. (I am an Instructure employee.) I expect 1.2 to be out in a short while, so I'll circle back on this in a week.

PR #648, opened 2015-06-19
Fix Canvadocs preview fails on first hit

shawniverson:
Add code to respect throttling and Retry_After header attribute of response.

### DIFF
--- a/lib/canvadocs.rb
+++ b/lib/canvadocs.rb
@@ -114,13 +114,27 @@ module Canvadocs
     #
     # Returns the json parsed response body of the call
     def api_call(method, endpoint, params={})
-      # dispatch to the right method, with the full path (/api/v2 + endpoint)
-      request = self.send("format_#{method}", "#{@url.path}/#{endpoint}", params)
-      request["Authorization"] = "Token #{token}"
-      response = @http.request(request)
+      responsewait = true
+      while responsewait == true
+        # dispatch to the right method, with the full path (/api/v2 + endpoint)
+        request = self.send("format_#{method}", "#{@url.path}/#{endpoint}", params)
+        request["Authorization"] = "Token #{token}"
+        response = @http.request(request)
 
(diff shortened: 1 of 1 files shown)


claydiffrient:
@shawniverson Thanks for the pull request.  I've got it started down the proper path to get it taken care of.  It's currently running through our test suite.  I'll get it in front of the proper engineers as well.  However, I don't see a CLA on file for you.  You can find it at https://github.com/instructure/canvas-lms/wiki/ica.pdf. Fill that out and send it back to me at clay@instructure.com

lib/canvadocs.rb
```diff
@@ -114,13 +114,27 @@ def view(session_id)
...
+        # dispatch to the right method, with the full path (/api/v2 + endpoint)
+        request = self.send("format_#{method}", "#{@url.path}/#{endpoint}", params)
+        request["Authorization"] = "Token #{token}"
+        response = @http.request(request)
+
+        case response.code
+        # Accepted and in queue
+        when /\A202/
+          retryafter = response['Retry-After'].to_i
+          sleep retryafter == 0 ? 2 : retryafter
```
simonista: Hey @shawniverson, thanks for the pull request.  Unfortunately, I don't think this is the right solution.  Adding a loop with `sleep` would bog down the controllers as they waited for the document to become available, and we actually wrap our calls to canvadocs in a timeout that defaults to 15 seconds, so it would often produce timeout errors instead of the document you are after.

I think instead we'd want to return the 202 or 429 out of the api library, and pass it all the way through back as the controller response, and make the javascript client code smart enough to look for non 302 redirects and handle the retry logic appropriately.   That's a bigger change, since right now the assumption with the api library is that it is either successful or throws an error.

simonista:
I'm going to close this for now, since it's not an approach we want to take, but feel free to re-open or open a new PS if you decide to work on an alternate solution.  Sorry :(

PR #650, opened 2015-06-25
Ignore RVM files

listrophy:
I believe that these files should either:
a) be gitignored to discourage specififying a ruby management system, or
b) added to the repo for ease of spin-up

As RVM tends to be a contentious issue, I assumed the core team would
want to choose "a".

(Signed ICA was sent ~1 minute ago)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,6 +12,8 @@
 
 .bundle/
 /.yardoc/
+/.ruby-gemset
+/.ruby-version
 /app/coffeescripts/ember/*/main.coffee
 /app/views/info/styleguide.html.erb
 /config/*.yml


claydiffrient:
@listrophy Thanks for the PR.  Unfortunately, we encourage users to ignore those files themselves rather than add them to the `.gitignore` for everyone.  You'll notice the guidelines for that [here](https://github.com/instructure/canvas-lms/blob/master/.gitignore#L1-L11) in the .gitignore file.

This prevents the .gitignore from getting out of control adding every niche version management tool, IDE, custom script, etc.

Again, thanks for submitting this, we look forward to other PRs from you in the future.

listrophy:
Ha. Sorry... my eyes totally skipped over that section.

I've got a routes.rb cleanup PR in the works. Won't be huge, but should add some consistency.

PR #691, opened 2015-09-11
Adding HTML5 video capability.

mattminuti:
Use standard HTML5 video tags without the source tag getting stripped away! Now you can upload a video file and reference its download link in the source field, like so:
`<video controls="controls" width="320" height="240"><source src="https://canvas.einsteinsworkshop.com/files/27/download?download_frd=1" type="video/mp4" /> </video>`
It would be nice to be able to do relative links, especially for exporting course pages, but I couldn't really see how to quickly do it.


ccutrer:
@mattminuti: thanks for pull request. Before we can continue with the process, we'll need a signed CLA from you, as described at https://github.com/instructure/canvas-lms/wiki#contributing. If you can send it and send a copy to opensource@instructure.com, we'll get this moving along quickly.

mattminuti:
CLA signed and sent!

simonista:
Sorry to be slow on this, I'll make sure this is run through our test suite and ping the appropriate team to check it out.

PR #704, opened 2015-10-09
Expose delayed jobs as an API route

grahamb:
This patch exposes delayed jobs data as an API route, which makes it accessible using a bearer token. Currently, a token is not accepted as valid authentication, and an Unauthorized error is returned to the client.

Test plan:

- create a user in the Site Admin account and grant it admin rights to Site Admin
- create a user in the Site Admin account (or any other account) but do not grant it admin rights to Site Admin
- create some delayed jobs, failed and otherwise
- test the following routes:
  - `GET /api/v1/jobs?only=jobs&flavor=[failed|future|current|waiting]`: should return list of jobs for specified flavor in JSON
  - `GET /api/v1/jobs/:id`: should return a specific non-failed job
  - `GET /api/v1/jobs/:id`: should return a specific failed job
  - `POST /api/v1/jobs/batch_update` with body `update_action=destroy&job_ids[]=$id1&job_ids[]=$id2`: (replacing
$id1 and $id2 with the IDs of real non-failed jobs) should return a success message with count of destroyed jobs.
  - `POST /api/v1/jobs/batch_update` with body
     `update_action=destroy&job_ids[]=$id1&job_ids[]=$id2&flavor=failed`:
(replacing $id1 and $id2 with the IDs of real failed jobs) will fail
pending instructure/canvas-jobs#2 (with that pull, it will return a
success message with count of destroyed jobs).
  - `POST /api/v1/jobs/batch_update` with body `update_action=destroy&flavor=future`: should return a success message with count of destroyed jobs
  - `POST /api/v1/jobs/batch_update` with body
    `update_action=hold&flavor=future`: should return an error report
(can't hold failed jobs).

When testing the routes, using the second user created (without Site Admin rights) should throw an Unauthorized error. Using a developer key should throw an Invalid Token error.

A signed Instructure Contributor License Agreement is on file under my employer, Simon Fraser University.

### DIFF
--- a/app/controllers/jobs_controller.rb
+++ b/app/controllers/jobs_controller.rb
@@ -23,7 +23,7 @@ class JobsController < ApplicationController
           @job_tags_refresh_seconds  = Setting.get('job_tags_refresh_seconds', 10.seconds.to_s).to_f
         end
 
-        format.js do
+        format.json do
           case params[:only]
           when 'running'
             render :json => {running: Delayed::Job.running_jobs.map{ |j| j.as_json(include_root: false, except: [:handler, :last_error]) }}
@@ -53,6 +53,7 @@ class JobsController < ApplicationController
 
     if params[:job_ids].present?
       opts[:ids] = params[:job_ids]
+      opts[:flavor] = params[:flavor] if params[:flavor] == 'failed'
     elsif params[:flavor].present?
       opts[:flavor] = params[:flavor]
       opts[:query] = params[:q]
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1743,6 +1743,18 @@ CanvasRails::Application.routes.draw do
     end
   end
 
+  # this is not a "normal" api endpoint in the sense that it is not documented or
+    # generally available to hosted customers. it also does not respect the normal
+    # pagination options; however, jobs_controller already accepts `limit` and `offset`
+    # paramaters and defines a sane default limit
+    ApiRouteSet::V1.draw(self) do
+      scope(controller: :jobs) do
+        get 'jobs', action: :index
+        get 'jobs/:id', action: :show
+        post 'jobs/batch_update', action: :batch_update
+      end
+    end
+
   # this is not a "normal" api endpoint in the sense that it is not documented
   # or called directly, it's used as the redirect in the file upload process
   # for local files. it also doesn't use the normal oauth authentication


config/routes.rb
```diff
@@ -1743,6 +1743,18 @@
     end
   end
 
+  # this is not a "normal" api endpoint in the sense that it is not documented or
+    # generally available to hosted customers. it also does not respect the normal
+    # pagination options; however, jobs_controller already accepts `limit` and `offset`
+    # paramaters and defines sane a default limit
```
ccutrer: s/sane a/a sane/

app/controllers/jobs_controller.rb
```diff
@@ -43,7 +43,7 @@ def show
     if params[:flavor] == 'failed'
       job = Delayed::Job::Failed.find(params[:id])
     else
-      job = Delayed::Job.find(params[:id])
+      job = Delayed::Job::Failed.find(params[:id]) rescue Delayed::Job.find(params[:id])
```
ccutrer: ewww rescue.  instead you can do

`Delayed::Job::Failed.find(Array.wrap(params[:id])).first ||
 Delayed::Job.find(params[:id])`

by passing an array, it will return an empty array if it's not found

I'm also concerned about this change in general, since regular jobs and failed jobs in the AR backend occupy different id namespaces (i.e you can have a job with id 1, and a failed job with id 1; this would effectively hide the non-failed one).

config/routes.rb
grahamb: I blame DayQuil for this one.

app/controllers/jobs_controller.rb
grahamb: Thanks for the pointer on `.wrap`. Good point on the potential ID collision. I think it's best if I just revert out that change and keep the original behaviour.

simonista:
I will pull this in and run it through our test suite

grahamb:
I'm still working out how to best handle batch updating failed jobs both here and in instructure/canvas-jobs. It seems like the most straightforward way would be to require `flavor=failed` when trying to update a failed job.

You have no idea how hard it is for me to resist modding this to also accept `flavour` as a parameter.

simonista:
So should I hold off on this for now, and wait for your next update?

grahamb:
Hold off for now; I'll ping you when I push an update.

grahamb:
@ccutrer & @simonista I've reworked this one a bit to be less ambigious when dealing with failed jobs. The bulk of the changes are really in instructure/canvas-jobs#2.

grahamb:
@simonista instructure/canvas-jobs#2 was just merged in, so this is now ready to get pulled in for testing.

app/controllers/jobs_controller.rb
```diff
@@ -53,6 +53,7 @@ def batch_update
 
     if params[:job_ids].present?
       opts[:ids] = params[:job_ids]
+      opts[:flavor] = params[:flavor] if params[:flavor].present? && params[:flavor] == 'failed'
```
ccutrer: ditto on the `present?` here

grahamb: Fixed with 1f36126 (squashed)

ccutrer:
:+1: , but it depends on https://gerrit.instructure.com/#/c/65733/

grahamb:
Can't see the gerrit link - I'm assuming that's running through the test suite?

ccutrer:
@grahamb: it's a commit to bump canvas to the freshly release canvas-jobs gem

grahamb:
@simonista any update on this?

simonista:
Finally passed the test suite, I'll see what we can do about getting it in today.

grahamb:
Any update on getting this pulled in?

simonista:
Yeah, I guess 17 days ago didn't happen :(.  QA just needs to do a run-through with it, they've just been super busy.

grahamb:
Thanks - just wanted to update our internal tracking for this one.

PR #728, opened 2015-11-25
Local file storage only: Fix IE file upload bug by nulling Content-Type header in GET request.

patchin:
File upload from IE (all versions) to local file storage fails because, under some conditions, IE will include a Content-Type header in a GET request, in response to a 302 redirect. When Rack receives the GET request and sees a non-null Content-Type header it tries to read a request body, which is missing. The request fails and the multistep file upload process fails.

This 302 redirect is a part of the file upload protocol described in the Canvas API Documentation (https://canvas.instructure.com/doc/api/file.file_uploads.html). As part of the file upload process the client code in IE initially sends a POST request and Canvas responds with a 302 redirect. The client code doesn't have a chance to intervene when the 302 redirect is received. IE responds to the 302 with a GET request, retaining the Content-Type from the initial POST request.

The fix is the addition of Rack middleware which will "nullify" the Content-Type header on a GET request to the file upload URL. So it's very specific when this action takes place and shouldn't be a performance problem.

Note that submissions to assignments are not affect as that uses another method of uploading files.

Test plan:

On a Canvas installation that uses local file storage (not s3):
- Upload a file to the Files area of a course or personal account. 
- File upload should succeed.

_NOTE: We have a signed CLA for all of Simon Fraser University (SFU) on file._

### DIFF
--- a/app/middleware/strip_content_type_from_get.rb
+++ b/app/middleware/strip_content_type_from_get.rb
@@ -1,0 +1,14 @@
+class StripContentTypeFromGet 
+  def initialize(app)
+    @app = app
+  end
+
+  def call(env)
+    if env['REQUEST_METHOD'] == "GET" && !env['CONTENT_TYPE'].nil?
+      if /^\/api\/v1\/files\/[0-9]+\/create_success$/ =~ env['REQUEST_PATH']
+        env['CONTENT_TYPE'] = nil
+      end
+    end
+    @app.call(env)
+  end
+end
--- a/config/application.rb
+++ b/config/application.rb
@@ -122,6 +122,7 @@ module CanvasRails
       app.config.middleware.insert_after(config.session_store, 'RequestContextSession')
       app.config.middleware.insert_before('ActionDispatch::ParamsParser', 'Canvas::RequestThrottle')
       app.config.middleware.insert_before('Rack::MethodOverride', 'PreventNonMultipartParse')
+      app.config.middleware.insert_before('PreventNonMultipartParse', 'StripContentTypeFromGet')
     end
 
     config.to_prepare do


FreedomBen:
Thanks @patchin!  Does this _only_ happen with IE then?

FreedomBen:
@patchin disregard that last question.  I had it pre-typed and then read your whole comment but accidentally hit "Comment" :facepalm:

patchin:
No problem! :-) Yes, I could only reproduce the issue with IE. I couldn't reproduce it with Chrome and Firefox on Windows. Also I couldn't reproduce the issue with Chrome, Firefox and Safari on OSX.

codekitchen:
Hm I'm curious why this would only affect local files, as s3 uploads also do a redirect after the upload completes. I'm confident in saying it doesn't affect s3 uploads, because we'd be hearing a ton about it.

The only differences I can see are (a) the s3 redirect comes from a separate domain, the s3 domain, and (b) the s3 redirect is a 301 instead of a 302. Perhaps the IE bug is specific to 302 redirects, and we can just change the response code? Or there's another difference I haven't noticed.

codekitchen:
Correction: S3 used to use a 301, it now uses a 303

patchin:
Thanks, I'll see if changing the status code makes a difference.

patchin:
Changing the status code to 301,303 or 307 didn't make any difference.

codekitchen:
I don't like the idea of adding a new middleware just to work around this bug. I'm optimistic about there being a simpler solution, since the S3 upload path isn't hitting the same problem. I think we need to find out more about the root cause, either through further testing or searching for others who have hit the same problem with IE. Are you able to do that? Otherwise I can try to schedule it internally but it might be some time.

patchin:
There is a simpler solution that involves just changing FilesController#api_create. Instead of redirecting to api_v1_files_create_success_url just return the json at that url. But this would require rewriting documentation about the file upload process in the API. Here is a link to another person who's had the same problem with IE: http://stackoverflow.com/questions/25874529/rack-throwing-eoferror-bad-content-body. I've spent some time exploring options on the client and server side but they seemed to involve more disruptive code changes than adding the middleware, especially changes to the client side.

codekitchen:
I was able to dig into this a bit more, and found the relevant difference between s3 and local file uploads. The Canvas web UI uses an undocumented api parameter `no_redirect` to tell the file uploads API that it doesn't want to be redirected, which causes the preflight request to set `success_action_status: 201` in the upload params that are sent along with the file upload.

S3 reads this param and responds with a 201 instead of a 303, and then the web UI manually calls create_success. So IE never even gets a redirect from S3. However, the local files endpoint `FilesController#api_create` doesn't look for `success_action_status`, so it always redirects.

If we modify the `api_create` action to look for the `success_action_status` parameter and respond with 201 instead of redirecting when it's provided, I think that should solve the IE problem.

Whether this was the best way to work around this issue in the first place is another question, but seeing as how the undocumented parameter is always in place, I think it's fine to use it to solve this bug.

patchin:
@codekitchen Thanks for looking into it! The solution sounds fine to me. I don't know how I missed that no_redirect detail.

claydiffrient:
@patchin I'm assuming that the resolution you had back in December was acceptable.  I'm going to go ahead and close the PR at this point.  If that's a problem please let me know and I'll reopen it for consideration.

patchin:
@claydiffrient Hi, has @codekitchen incorporated his suggested fix already? Thanks

codekitchen:
I haven't done anything with this, no.

simonista:
I've converted this into an Issue: https://github.com/instructure/canvas-lms/issues/990, and am going to close this PR

PR #731, opened 2015-11-27
Performance tune activesupport-suspend_callbacks for Rails 3.2 and 4.0

dgynn:
This PR reduces the performance impact of the `activesupport-suspend_callbacks` gem in the common case where callbacks are not being suspended. Performance is also slightly improved when callbacks are suspended.

For Rails 4.0 on Ruby 2.2
```
Before:
User.new      4.725k (± 9.8%) i/s (iterations per second)
User.last     1.071k (± 3.4%) i/s
Courses#show page:  552207 bytes and 4132 objects allocated

After:
User.new      7.248k (± 2.0%) i/s - 50% improvement
User.last     1.227k (± 2.8%) i/s - 15% improvement
Courses#show page:   12640 bytes and  316 objects allocated - 90+% reduction
```

The measurements after tuning are close to the performance if the gem was not included. Improvements for Rails 3 are similar. 

Here is the benchmark script I used to do the comparisons: https://gist.github.com/dgynn/4cc827030314280b5170

The performance gains come primarily by falling back to the default ActiveSupport callback logic if callbacks are not currently suspended. The check for whether callbacks are suspended has been refactored to reduce object creation. Only when callbacks are suspended and they are suspended in a way that affects the specific callback is a new callback chain compiled.

Test Plan:
  - Run the specs for the gem and the full test suite (for Rails 3 and 4)
  - Compare performance in a staging environment

This is my first PR so let me know what else would be helpful. CLA is on file.

### DIFF
--- a/gems/activesupport-suspend_callbacks/lib/active_support/callbacks/suspension.rb
+++ b/gems/activesupport-suspend_callbacks/lib/active_support/callbacks/suspension.rb
@@ -72,9 +72,10 @@ module ActiveSupport::Callbacks
     #   end
     #
     def suspended_callback?(callback, kind, type=nil)
-      suspended_callbacks.include?(callback, kind, type) ||
+      instance_variable_defined?(:@suspended_callbacks) &&
+        suspended_callbacks.include?(callback, kind, type) ||
       suspended_callback_ancestor &&
-      suspended_callback_ancestor.suspended_callback?(callback, kind, type)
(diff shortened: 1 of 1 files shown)


FreedomBen:
@dgynn Just so you know, this PR got a "That PR sounds awesome" internally.  Thanks for your contribution :+1:

dgynn:
@FreedomBen Fantastic! Thanks for the speedy merge.

PR #746, opened 2016-01-06
Updating es.yml to call attention to a few translation errors

danielmcq:
I understand that the language translation YML files are generated, but I noticed a few inconsistencies in the es translation, so I'm doing this pull request so they get visibility.

### DIFF
--- a/config/locales/es.yml
+++ b/config/locales/es.yml
@@ -2,12 +2,12 @@
 # Do not edit it by hand, your changes will be overwritten.
 --- 
 es: 
-  1_item_scores_5512cd03: "2- Puntuaciones de los elementos"
+  1_item_scores_5512cd03: "1- Puntuaciones de los elementos"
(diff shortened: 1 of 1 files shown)


claydiffrient:
@danielmcq Thanks for the PR.  I'm going to send these suggestions to our translators.  But since we can't actually merge onto the auto-generated file, I'm going to close the PR.

simonista:
Also, if you're interested in fixing the issue at the source, you can create an account on transifex (https://www.transifex.com/projects/p/canvas-lms/resource/canvas-lms/) and suggest fixes there directly.

danielmcq:
Thanks. I'll head over to that page and make the suggestions.

claydiffrient:
@danielmcq I'm told that those translations have been updated in Transifex and will be replicated into the Canvas codebase.

PR #625, opened 2015-05-05
Add 'mark done' module progression constraint (Take 5)

kjellm:
In module based courses, to make it easier for students to keep
track of their progress. This commit adds support for 'mark as
done' requirements to the module progressions. This to better
cater to students who needs multiple reading sessions before they
feel they are done with a module item (i.e wiki pages).

This requirement can also be used on assignments that are purely
for self studying.

Test plan:
  - In a course with modules
  - Make sure there is one wiki page and one assignment associated with a module
  - Add 'mark as done' as module completion requirements to both module items
  - As a student of the course, go to the modules page
  - Oberve that both module items are labeled as "must mark as done"
  - Visit the module items each in turn and click the "mark as done" button.
  - On the modules page, oberve that both module items are now labeled as "marked done".

PR #751, opened 2016-01-26
Repaired mark as done button in production.

Kulgar:
The button was working fine in development but not in production.
The events was never triggered (confirmed that by adding a 'console.log').
I had to modify the click listener with a 'on' one to make it work.
It is quite logic I think, as the button is actually 'appended' using javascript.
In production it was failing both for firefox and chrome.

Test plan:
* Create a course, with a module and a page (and/or an assignment)
* Add a mark as done requirement for the newly created resource
* Invite a student and connect using his account
* Go to the page (of course you did not forget to publish it?)
* Try the mark as done button

Again, it was working fine in development but not in production. I do wonder why... if you have any ideas, let me know.

The code I modified was originally added here: #625

### DIFF
--- a/app/coffeescripts/bundles/assignment_show.coffee
+++ b/app/coffeescripts/bundles/assignment_show.coffee
@@ -12,8 +12,9 @@ require [
 ], (INST, I18n, $, Assignment, PublishButtonView, SpeedgraderLinkView, vddTooltip, MarkAsDone) ->
 
   $ ->
-    $('#mark-as-done-checkbox').click ->
+    $('#content').on('click', '#mark-as-done-checkbox', ->
       MarkAsDone.toggle(this)
+    )
 
   $ ->
     $el = $('#assignment_publish_button')
--- a/app/coffeescripts/bundles/wiki_page_show.coffee
+++ b/app/coffeescripts/bundles/wiki_page_show.coffee
@@ -7,8 +7,9 @@ require [
 ], ($, WikiPage, WikiPageView, MarkAsDone) ->
 
   $ ->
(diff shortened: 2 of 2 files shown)


claydiffrient:
@Kulgar thanks for the pull request.  If you wouldn't mind squashing the two commits that would be great!

Kulgar:
@claydiffrient : And... done :wink:

claydiffrient:
@Kulgar Great!  I'm getting this running through our test suite now.  I'll keep the status of this pull request updated so you know what's going on.

One small thing that needs to change though is the commit message.  If you could put a blank line between the first two lines of the message that would be very helpful :)  I'm sorry I didn't notice that before.

Kulgar:
@claydiffrient : And... done again :wink: 

When you say: I'm getting this running through our test suite now. 
Is this something I can do (like rspec or something)? Or is it something you have to do on your side only?

claydiffrient:
@Kulgar So it's really helpful if you check at least the relevant specs locally via rspec or the javascript test suite.

Ruby: `bundle exec rspec path/to/test/file.rb`
JS: `JS_SPEC_MATCHER="**/nameofspecfile.js" bundle exec rake js:test`

The trouble is the test suite is fairly large and will take a long time to run in a non-cloud or parallelized environment. So we run all the tests in the cloud and parallelized on our end so it gets done in ~45 minutes, barring any potential problems.  The other thing that we also run is a full suite against proprietary plugins as well so that we know it runs in our hosted instances as well.

So yeah, we have to run it, but you can ensure most things work by doing at least a partial run on your end first :)

PR #768, opened 2016-02-11
Support class reloading with BroadcastPolicy

dgynn:
Since Notification and NotificationPolicy can be reloaded, BroadcastPolicy needs to re-cache the references on reload. This problem was discussed on the mailing list on 11/23/15.

To test:
You can experience a reloading error before applying this patch by triggering a notification after a reload.
```
rails console
reload!
Course.last.enroll_student(User.last).accept!
```
This will cause `ArgumentError: A copy of Notifier has been removed from the module tree but is still active!`

With this patch applied, there should be no error (assuming that student can be enrolled in that class).

### DIFF
--- a/config/initializers/broadcast_policy.rb
+++ b/config/initializers/broadcast_policy.rb
@@ -1,4 +1,6 @@
 require 'broadcast_policy'
-BroadcastPolicy.notifier = lambda { Notifier.new }
-BroadcastPolicy.notification_finder = lambda { NotificationFinder.new(Notification.all_cached) }
+Rails.configuration.to_prepare do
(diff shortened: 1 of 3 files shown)


ccutrer:
I'm not able to repo the error locally, but the code makes sense, and I am able to confirm with:

```
rails c
BroadcastPolicy.notifier
reload!
BroadcastPolicy.notifier.class.object_id != Notifier.object_id
```

so :+1:, we'll get it running through CI

dgynn:
Super. Thanks @ccutrer. And thanks for the switchman merges too.

PR #769, opened 2016-02-11
Remove Thor dependency from moodle_importer

dgynn:
The `moodle_importer` does not directly use `thor` and has no need for a specific version dependency. `moodle2cc` already has an unversioned thor dependency and `railties` itself requires `>= 0.18.1`.

Specifically for me, this hard dependency caused issues with the current releases of  `derailed_benchmarks` and `foreman` which required `~> 0.19` and `~> 0.19.1` respectively. 

0.19.1 is the current release as of 3/2014.

Test Plan:
- run bundle update
- run `moodle2cc help` to verify moodle2cc thor tasks still load

### DIFF
--- a/gems/plugins/moodle_importer/moodle_importer.gemspec
+++ b/gems/plugins/moodle_importer/moodle_importer.gemspec
@@ -18,5 +18,4 @@ Gem::Specification.new do |s|
 
   s.add_dependency "rails", ">= 3.2", "< 4.3"
   s.add_dependency "moodle2cc", "0.2.28"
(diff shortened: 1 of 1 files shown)


ccutrer:
:+1: 

Note that normally we like to keep gems "locked" to a specific version (though in the Canvas gemfile, not a gemspec of a private gem) - long story but we can't commit the Gemfile.lock.  But in this case, since it's only a gem for command line handling of something not directly run by Canvas, it's fine.

PR #771, opened 2016-02-12
Reduce unnecessary activity stream XHR fetches

dgynn:
On a the dashboard page for a user with multiple courses, duplicate XHR fetches are sent. For example with 12 cards shown, there will be 78 fetches (12+11+10...+1) rather than 12.

Since all of the DashboardCard React views share a single CourseActivitySummaryStore, any update to the store triggers every card to request an update from the store and if it does not exist, it fires another XHR fetch.

This PR prevents the additional fetches by storing an empty result `{}` in the store which will be replaced when the XHR call completes.

Test Plan
- Load the dashboard page for a user multiple course cards
- Observe the XHR requests for `/api/v1/courses/*/activity_stream/summary`
- There should only be 1 request per card and unread activity should show up


claydiffrient:
Merged and closed with 390b7a2c1399fe178a66901259e120401da6e5c5

PR #781, opened 2016-03-06
add missing section update API doc parameters

cmurtaugh:
The API doc comments on the section update method are missing parameters, resulting in incomplete swagger spec and API documentation.

This PR adds documentation for the following parameters (copied from the section create method doc):
```
course_section[name]
course_section[sis_section_id]
course_section[start_at]
course_section[end_at]
course_section[restrict_enrollments_to_section_dates]
```

### DIFF
--- a/app/controllers/sections_controller.rb
+++ b/app/controllers/sections_controller.rb
@@ -228,7 +228,22 @@ class SectionsController < ApplicationController
   end
 
   # @API Edit a section
-  # Modify an existing section.  See the documentation for {api:SectionsController#create create API action}.
+  # Modify an existing section.
+  #
+  # @argument course_section[name] [String]
+  #   The name of the section
+  #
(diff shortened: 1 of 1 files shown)


app/controllers/sections_controller.rb
```diff
@@ -230,6 +230,21 @@ def uncrosslist
   # @API Edit a section
   # Modify an existing section.  See the documentation for {api:SectionsController#create create API action}.
```
ccutrer: since the parameters are copied now, there's no longer any need to refer to the Create documentation

cmurtaugh: Good point - fixed.

cmurtaugh:
I squashed these commits; let me know if anything else is needed.

ccutrer:
:+1:

PR #782, opened 2016-03-06
fix course update API doc params

cmurtaugh:
The course update method's API docs are missing the `course[is_public_to_auth_users]` param, resulting in an incomplete swagger spec.  This PR adds documentation of this parameter and fixes a minor typo in the `course[is_public]` documentation.

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -518,10 +518,10 @@ class CoursesController < ApplicationController
   #   - 'public_domain' (Public Domain).
   #
   # @argument course[is_public] [Boolean]
-  #   Set to true if course if public.
+  #   Set to true if course is public to both authenticated and unauthenticated users.
   #
   # @argument course[is_public_to_auth_users] [Boolean]
-  #   Set to true if course if public to authenticated users.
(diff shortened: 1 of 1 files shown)


app/controllers/courses_controller.rb
```diff
@@ -1926,7 +1926,10 @@ def copy_course
   #   - 'public_domain' (Public Domain).
   #
   # @argument course[is_public] [Boolean]
-  #   Set to true if course if public.
+  #   Set to true if course is public.
+  #
+  # @argument course[is_public_to_auth_users] [Boolean]
```
ccutrer: if we're going to document this, we should return it in the API response for courses, and clarify that is_public makes it public to both authenticated and non-authenticated users

cmurtaugh: Good point - fixed.

cmurtaugh: The API response for courses already includes `is_public_to_auth_users`; I've added some doc to clarify the difference between `is_public` and `is_public_to_auth_users`.

cmurtaugh:
I squashed these commits; let me know if anything else is needed.

ccutrer:
:+1:

PR #790, opened 2016-03-18
Change the max height for lti.frameResize.

rutger1413:
The max height being 5000 is far to short for some of our content. I see the need for setting a max height, but having it set to a max of 5000 does not allow for displaying large content pieces. I guess the question is, where do you stop.

### DIFF
--- a/public/javascripts/tool_inline.js
+++ b/public/javascripts/tool_inline.js
@@ -148,7 +148,7 @@ window.addEventListener('message', function(e) {
     switch (message.subject) {
       case 'lti.frameResize':
         var height = message.height;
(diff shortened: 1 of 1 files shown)


claydiffrient:
@rutger1413 After consultation with several different teams here at Instructure, we think that limitation should just be removed completely.  If you could modify your PR to do that we'd gladly accept it.

The other thing is that your PR is based on the `stable` branch, but it would be preferable to be based on the `master` branch instead.

rutger1413:
I couldn't get the file changed on the master branch to be added to this pull request, so I created another pull request:
https://github.com/instructure/canvas-lms/pull/803

PR #799, opened 2016-03-30
Permit data-* attributes in HTML content

dgynn:
This PR allows data-* attributes in HTML to pass the santization process.
For example, this allows integration with Embedly if the content includes HTML such as...
```html
<a class="embedly-card" href="http://www.instructure.com/" data-card-controls="0" data-card-type="article">Instructure</a>
```
...and the appropriate JavaScript is loaded.

Test Plan:
- Edit a wiki page using the HTML editor and enter custom HTML5 code that includes data attributes
- Verify the attributes exist in the HTML when the page is rendered

Note: This is minor version upgrade to sanitize. It does not upgrade to 3.x or 4.x which could introduce more changes.

### DIFF
--- a/gems/canvas_sanitize/lib/canvas_sanitize/canvas_sanitize.rb
+++ b/gems/canvas_sanitize/lib/canvas_sanitize/canvas_sanitize.rb
@@ -90,6 +90,7 @@ module CanvasSanitize #:nodoc:
                    'role',
                    'lang',
                    'dir',
+                   :data,  # Note: the symbol :data allows for arbitrary HTML5 data-* attributes
                    'aria-labelledby',
                    'aria-atomic',
                    'aria-busy',
--- a/Gemfile.d/app.rb
+++ b/Gemfile.d/app.rb
@@ -92,7 +92,7 @@ gem 'ruby-saml-mod', '0.2.7'
 gem 'rubycas-client', '2.3.9', require: false
 gem 'rubyzip', '1.1.1', require: 'zip'
 gem 'safe_yaml', '1.0.4', require: false
-gem 'sanitize', '2.0.6', require: false
+gem 'sanitize', '2.1.0', require: false
 gem 'shackles', '1.1.0'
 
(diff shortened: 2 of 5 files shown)


gems/canvas_sanitize/lib/canvas_sanitize/canvas_sanitize.rb
```diff
@@ -90,6 +90,7 @@ def self.clean_style_value(config, value)
                    'role',
                    'lang',
                    'dir',
+                   :data,
```
claydiffrient: @dgynn It'd be nice if this maintained consistency with the surrounding entries in the array, i. e. a string rather than a symbol.

dgynn: Thanks for taking a look, @claydiffrient. The sanitize gem uses `:data` as a symbol to differentiate it from an attribute named 'data'.

> To allow arbitrary HTML5 data-* attributes, use the symbol :data in place of an attribute name.

claydiffrient: Would you mind adding a comment there indicating that?  It would make it easier for those looking in the future to understand the reasoning.

dgynn: @claydiffrient I've updated the code with a comment about what `:data` does and mentioned that it should be a symbol. Let me know if that should be more clear.

claydiffrient:
@dgynn Sorry for the delay on this.  I've got it running through our test suite now.  One more thing, if you wouldn't mind squashing the two commits to one, that would help to keep the git history a bit cleaner.  I would love to just use the squash and merge feature of GitHub, but we don't merge directly from GitHub, it first goes through our internal Gerrit repository and gets replicated.

dgynn:
@claydiffrient No problem. I've squashed the commits. Thanks.

PR #815, opened 2016-04-27
VeriCite plugin integration into Assignments, Gradebook, and Speedgrader

baholladay:
I have added VeriCite to the list of "/plugins". This is how you enable VeriCite for your account. Once enabled, an option will show up in the assignment settings for "inline" or "upload" assignment types. This plugin uses all of the existing turnitin data structures in the tables as requested. The only additional DDL required is a flag on the assignments table in order to differentiate which service is enabled:

alter table assignments ADD vericite_enabled boolean;

Here is a video of the plugin in action:

https://youtu.be/ttdFkyR1VTM?t=36s (note that it now uses the plugin configuration instead of the account settings shown in the video)


app/coffeescripts/gradebook2/GRADEBOOK_TRANSLATIONS.coffee
```diff
@@ -14,4 +14,5 @@ define ['i18n!gradebook2'], (I18n) ->
     submission_tooltip_media_recording: I18n.t('titles.media_recording', "Media Recording Submission"),
     submission_tooltip_online_quiz: I18n.t('title.quiz', "Quiz Submission"),
     submission_tooltip_turnitin: I18n.t('title.turnitin', 'Has Turnitin score'),
+    submission_tooltip_vericite: I18n.t('title.vericite', 'Has VeriCite score'),
```
ktgeek: This line and its preceding line should be collapsed into a plagiarism score.

app/coffeescripts/gradebook2/SubmissionCell.coffee
```diff
@@ -105,9 +106,12 @@ define [
       innerContents = null if opts.submission.workflow_state == 'pending_review' && !isNaN(innerContents)
       innerContents ?= if submission_type then SubmissionCell.submissionIcon(submission_type) else '-'
 
-      if turnitin = extractData(opts.submission)
+      if turnitin = extractDataTurnitin(opts.submission)
         specialClasses.push('turnitin')
         innerContents += "<span class='gradebook-cell-turnitin #{htmlEscape turnitin.state}-score' />"
+      if vericite = extractDataVeriCite(opts.submission)
+        specialClasses.push('vericite')
+        innerContents += "<span class='gradebook-cell-vericite #{htmlEscape vericite.state}-score' />"
```
ktgeek: This should be collapsed into a single span with the turnitin item above.

app/coffeescripts/gradebook2/Turnitin.coffee
```diff
@@ -25,7 +25,7 @@ define [
       data.state = stateList[max(states)]
       data
 
-    extractDataFor: (submission, key, urlPrefix) ->
+    extractDataForTurnitin: (submission, key, urlPrefix) ->
```
ktgeek: This should be named more generically and either take a switch or some other augment to switch between "Turnitin" and "Vericite" modes

app/coffeescripts/gradebook2/VeriCite.coffee
```diff
@@ -0,0 +1,36 @@
+define [
```
ktgeek: This file doesn't need to exist.  See the notes in Turnitin.coffee

app/coffeescripts/gradebook2/Turnitin.coffee
```diff
@@ -4,7 +4,7 @@ define [
 ], (I18n, {max, invert}) ->
 
   Turnitin =
-    extractData: (submission) ->
+    extractDataTurnitin: (submission) ->
```
ktgeek: This should be renamed to be more generic (and provide a switch if needed.)

app/coffeescripts/models/Assignment.coffee
```diff
@@ -29,6 +30,9 @@ define [
       if (turnitin_settings = @get('turnitin_settings'))?
         @set 'turnitin_settings', new TurnitinSettings(turnitin_settings),
           silent: true
+      if (vericite_settings = @get('vericite_settings'))?
+        @set 'vericite_settings', new VeriCiteSettings(vericite_settings),
```
ktgeek: Is this a redundant/second field in the database?  We'd like to see the turnitin_settings reused with something like a "provider" key.

app/coffeescripts/models/Assignment.coffee
```diff
@@ -356,6 +372,8 @@ define [
         data.assignment_overrides = new AssignmentOverrideCollection overrides
       if (turnitin_settings = data.turnitin_settings)?
         data.turnitin_settings = new TurnitinSettings turnitin_settings
+      if (vericite_settings = data.vericite_settings)?
+        data.vericite_settings = new VeriCiteSettings vericite_settings
```
ktgeek: Needed?  Should this be generalized?

app/coffeescripts/views/assignments/EditView.coffee
```diff
@@ -193,6 +201,14 @@ AssignmentGroupSelector, GroupCategorySelector, toggleAccessibly, RCEKeyboardSho
         turnitinDialog.off()
         turnitinDialog.remove()
 
+    showVeriCiteDialog: (ev) =>
+      ev.preventDefault()
+      vericiteDialog = new VeriCiteSettingsDialog(model: @assignment.get('vericite_settings'))
```
ktgeek: Combine/change to only come from one place on the assignment.

app/coffeescripts/views/assignments/VeriCiteSettingsDialog.coffee
```diff
@@ -0,0 +1,68 @@
+define [
```
ktgeek: Require both templates from the turnitin settings dialog (renamed generically) and render the right form.  This file is only 2 lines different from the turnitin one, and one of those lines is the class name.

app/controllers/submissions_controller.rb
```diff
@@ -491,6 +491,46 @@ def resubmit_to_turnitin
     end
   end
 
+   def vericite_report
```
ktgeek: We don't mind the separate route, but since this function and the turnitin version are 99% the same, a combined method seems to be more generically going forward.

app/controllers/submissions_controller.rb
```diff
@@ -491,6 +491,46 @@ def resubmit_to_turnitin
...
+      if url
+        redirect_to url
+      else
+        flash[:notice] = t('errors.no_report', "Couldn't find a report for that submission item")
+        redirect_to named_context_url(@context, :context_assignment_submission_url, @assignment.id, @submission.user_id)
+      end
+    end
+  end
+
+  def resubmit_to_vericite
```
ktgeek: Similar to above, this code is largely the same so should be combined with the turn it in.

app/models/assignment.rb
```diff
@@ -399,6 +406,41 @@ def turnitin_settings=(settings)
...
+    end
+    self.save
+    return self.turnitin_settings[:current]
+  end
+
+  def vericite_settings
+    self.turnitin_settings(VeriCite::Client.default_assignment_vericite_settings)
+  end
+
+  def vericite_settings=(settings)
```
ktgeek: This should combine up with turnitin_settings and have an argument that is either the normalizer passed in or a switch for which client to use.

app/models/submission.rb
```diff
@@ -143,6 +143,7 @@ def needs_grading_changed?
   before_save :update_if_pending
   before_save :validate_single_submission, :infer_values
   before_save :prep_for_submitting_to_turnitin
+  before_save :prep_for_submitting_to_vericite
```
ktgeek: All of the vericite methods that have turnitin variants should be revisited.  Methods that are largely similar except for one or two lines should be rewritten to be generalized.

app/models/submission.rb
```diff
@@ -248,6 +250,20 @@ def new_version_needed?
       )
     }
     can :view_turnitin_report
+
+    given { |user, session|
```
ktgeek: This seems like it should be generic.  Maybe :view_plagiarism_report with combined logic and/or helper functions.

app/models/submission.rb
```diff
@@ -248,6 +250,20 @@ def new_version_needed?
...
+      (assignment.context.grants_right?(user, session, :manage_grades) ||
+        case assignment.vericite_settings[:originality_report_visibility]
+          when 'immediate'; true
+          when 'after_grading'; current_submission_graded?
+          when 'after_due_date'; assignment.due_at && assignment.due_at < Time.now.utc
+          when 'never'; false
+        end
+      )
+    }
+    can :view_vericite_report
```
ktgeek: This seems like it should be generic.  Maybe :view_plagiarism_report with combined logic and/or helper functions.

app/models/submission.rb
```diff
@@ -482,6 +498,268 @@ def turnitinable_by_lti?
     turnitin_data.select{|_, v| v.is_a?(Hash) && v.key?(:outcome_response)}.any?
   end
 
+  # VeriCite
+
+  # this function will check if the score needs to be updated and update/save the new score if so,
+  # otherwise, it just returns the vericite_data_hash
+  def vericite_data()
```
ktgeek: Parens not needed, style thing, but not breaking.

app/models/submission.rb
```diff
@@ -482,6 +498,268 @@ def turnitinable_by_lti?
...
+    if !self.readonly? && self.versions.current.number == self.version_number
+      self.vericite_data_hash.keys.each do |asset_string|
+        data = self.vericite_data_hash[asset_string]
+        next unless data && data.is_a?(Hash) && data[:object_id]
+        updateScores = updateScores || vericite_recheck_score(data)
+      end
+      # we have found at least one score that is stale, call VeriCite and save the results
+      if updateScores
+        check_vericite_status
+      end
```
ktgeek: This could potentially turn into one network call per submission on a large load like speed_grader_json.  This can't happen inline like this.  check_vericite_status should only ever happen when requested from the front-end or from a delayed job after the submissions.

app/models/submission.rb
```diff
@@ -482,6 +498,268 @@ def turnitinable_by_lti?
...
+    end
+    self.vericite_data_hash
+  end
+
+  def vericite_data_hash
+    # use the same backend structure to store "content review" data
+    self.turnitin_data
+  end
+
+  # this function looks at a vericite data object and determines whether the score needs to be rechecked (i.e. cache for 20 mins)
```
ktgeek: Given the large user base of canvas, in production this is a really bad idea, especially if its updating one submission at a time.

app/models/submission.rb
```diff
@@ -482,6 +498,268 @@ def turnitinable_by_lti?
...
+          updateScores = true
+        end
+      end
+    end
+    updateScores
+  end
+
+   VERICITE_STATUS_RETRY = 8
+
+  def check_vericite_status(attempt=1)
```
ktgeek: This appears to only work on the current submission.  Imagine a class of 1400 students being loaded into speedgrader, would this return in 1400 network calls to check the status?  This seems like it won't work at scale.

app/models/submission.rb
```diff
@@ -482,6 +498,268 @@ def turnitinable_by_lti?
...
+
+  def clear_vericite_errors(asset_data)
+    [:error_code, :error_message, :public_error_message].each do |key|
+      asset_data.delete(key)
+    end
+    asset_data
+  end
+
+
+  def resubmit_to_vericite
```
ktgeek: The submissions should either know directly or through the assignment which plagiarism service is in play and can call that.  We'd like to see a generic "resubmit_to_plagiarism_serivce" method.  (An alias should be created for resubmit_to_turnitin that points to this new generically named method.)

app/models/submission.rb
```diff
@@ -482,6 +498,268 @@ def turnitinable_by_lti?
...
+      assignment.vericite_enabled?
+  end
+
+  def vericite_data_changed!
+    @vericite_data_changed = true
+  end
+
+  def vericite_data_changed?
+    @vericite_data_changed
+  end
```
ktgeek: for lines 753-759: This shouldn't be any different from the turnitin methods.  Generizing with aliases should be considered.

app/models/submission.rb
```diff
@@ -663,6 +941,10 @@ def submission_history
       last_submitted_at = nil
       self.versions.sort_by(&:created_at).reverse_each do |version|
         model = version.model
+        # since vericite_data is a function, make sure you are cloning the most recent vericite_data_hash
+        if self.vericiteable?
+          model.turnitin_data = self.vericite_data
```
ktgeek: This could side effect into network calls when bringing up submission history.  This isn't a good idea here at scale.

app/presenters/grade_summary_assignment_presenter.rb
```diff
@@ -121,6 +121,15 @@ def turnitin
     t.try(:[], :state) ? t : nil
   end
 
+  def vericite
```
ktgeek: Other than pulling from the vericite_data method, this is the same as the turnitin item above.  These should be combined.

app/presenters/grade_summary_assignment_presenter.rb
```diff
@@ -138,7 +147,7 @@ def graph
   end
 
   def file
-    @file ||= submission.attachments.detect{|a| submission.turnitin_data && submission.turnitin_data[a.asset_string] }
+    @file ||= submission.attachments.detect{|a| ((submission.turnitin_data && submission.turnitin_data[a.asset_string]) || (submission.vericite_data && submission.vericite_data[a.asset_string])) }
```
ktgeek: This looks like it can be simplified/generalized.

app/stylesheets/bundles/speed_grader.scss
```diff
@@ -680,6 +680,12 @@ body .ui-selectmenu-dropdown:hover .ui-selectmenu-icon, body .ui-state-active .u
   padding: 3px 0;
 }
 
+.vericite_info {
```
ktgeek: No different from turnitin_info.  The turnitin_info should be reused.

app/stylesheets/bundles/speed_grader.scss
```diff
@@ -733,6 +739,20 @@ body .ui-selectmenu-dropdown:hover .ui-selectmenu-icon, body .ui-state-active .u
           @include fontSize(12px);
         }
       }
+      .vericite_score_container {
```
ktgeek: No different from turnitin's version.  The turnitin version should be reused.

app/models/submission.rb
baholladay: This call requests every single score for the entire assignment in one call and then cache's the results for the subsequent requests. This can only result in one network call for a page load. There is an additional safety measure that makes sure the score's cache is only called once every 5 minutes at a maximum for any assignment no matter which user requests it.

app/stylesheets/bundles/speed_grader.scss
```diff
@@ -733,6 +739,20 @@ body .ui-selectmenu-dropdown:hover .ui-selectmenu-icon, body .ui-state-active .u
...
+        position: absolute;
+        top: 0;
+        right: 20px;
+      }
+      .vericite_similarity_score {
+        font-size: 9px;
```
ktgeek: Why a different in font vs turnitin?  (Otherwise, same as turnitin's version and that should be used.)

app/models/submission.rb
baholladay: Same as above:
This call requests every single score for the entire assignment in one call and then cache's the results for the subsequent requests. This can only result in one network call for a page load. There is an additional safety measure that makes sure the score's cache is only called once every 5 minutes at a maximum for any assignment no matter which user requests it.

app/stylesheets/components/deprecated/_ui_listview.scss
```diff
@@ -62,6 +62,9 @@
       .turnitin_similarity_score {
         margin-right: 10px;
       }
+      .vericite_similarity_score {
```
ktgeek: same as turnitin above, that be reused

app/stylesheets/jst/SubmissionDetailsDialog.scss
```diff
@@ -35,6 +35,9 @@
     a.turnitin_similarity_score {
       @include fontSize(10px);
     }
+    a.vericite_similarity_score {
```
ktgeek: same as turnitin above, that be reused

app/stylesheets/pages/_vericite.scss
```diff
@@ -0,0 +1,98 @@
+@import "base/environment";
```
ktgeek: This file is the same as the turnitin other than variable renames.  This also in its current form potentially breaks some a11y color choices.
The turnitin file should just be used.

app/stylesheets/pages/assignments/_assignments.scss
```diff
@@ -66,6 +66,13 @@
   }
 }
 
+#assignment_vericite_settings {
+  min-width: 400px;
+  label {
+    width: 95%;
```
ktgeek: Seems very similar to turnitin except for width.  Could turnitin's be reused?

app/stylesheets/pages/gradebook2/_gradebook2.scss
```diff
@@ -418,6 +418,33 @@ $cell_height: 33px;
   }
 }
 
+.gradebook-cell-vericite {
```
ktgeek: The turnitin definition should be used, its the same.

app/stylesheets/pages/gradebook2/_gradebook2.scss
```diff
@@ -418,6 +418,33 @@ $cell_height: 33px;
...
+  &.warning-score {
+    background-image: url(/images/vericite_warning_score.png);
+  }
+  &.problem-score {
+    background-image: url(/images/vericite_problem_score.png);
+  }
+  &.failure-score {
+    background-image: url(/images/vericite_failure_score.png);
+  }
+}
```
ktgeek: Lines 428-445: the existing pngs should probably be used.  The only differences appear to be opacity?  If changes are desired a discussion with Instructure's designers should probably be had.

app/stylesheets/pages/react_gradebook/_react_gradebook.scss
```diff
@@ -219,6 +219,33 @@ $cell_height: 33px;
   }
 }
 
+.gradebook-cell-vericite {
```
ktgeek: The turnitin one should be reused, (see comments in the gradebook.scss)

app/views/assignments/_vericite.html.erb
```diff
@@ -0,0 +1,17 @@
+<%- if @context.vericite_comments.present? %>
+  <tr>
+    <td colspan="2" class='vericite-comments'>
+      <%= @context.vericite_comments %>
+    </td>
+  </tr>
+<%- end %>
+<%- if @context.vericite_pledge.present? && !@context.vericite_pledge.empty? %>
```
ktgeek: present? covers !empty?, this if should be simplified.

app/views/gradebooks/speed_grader.html.erb
```diff
@@ -158,11 +158,13 @@
             <div id="submission_files_list">
               <div id="submission_file_hidden"  class="submission-file" style='display:none;'>
                 <span class="turnitin_score_container"></span>
+                <span class="vericite_score_container"></span>
```
ktgeek: Since turnitin and vericite are mutually exclusive, let's use the same container for both.

app/views/gradebooks/speed_grader.html.erb
```diff
@@ -158,11 +158,13 @@
...
                 <a class="submission-file-download" title="<%= t('titles.download_file', "Download this file") %>" href="<%= context_url(@context, :context_assignment_submission_url, @assignment.id, "{{submissionId}}", :download => "{{attachmentId}}") %>"><%= t('links.download_file', "Download File") %></a>
                 <a class="display_name no-hover" href="<%= context_url(@context, :context_assignment_submission_url, @assignment.id, "{{submissionId}}", :download => "{{attachmentId}}", :inline => 1) %>" title="<%= t('titles.preview_file', "Preview this File") %>">???</a>
               </div>
             </div>
             <div class="turnitin_info_container"></div>
+            <div class="vericite_info_container"></div>
```
ktgeek: Since turnitin and vericite are mutually exclusive, let's use the same container for both.

app/views/gradebooks/speed_grader.html.erb
```diff
@@ -198,9 +200,11 @@
               <%= render partial: "grading_box_extended",
                          locals: { assignment: @assignment } %>
               <span class="turnitin_score_container"></span>
+              <span class="vericite_score_container"></span>
```
ktgeek: Since turnitin and vericite are mutually exclusive, let's use the same container for both.

app/views/gradebooks/speed_grader.html.erb
```diff
@@ -198,9 +200,11 @@
...
               <a href="#" id="submit_same_score"><br><%= t('links.submit_same_score', "Use this same grade for the resubmission") %></a>
               <a style="display: none;" title="POST" href="<%= context_url(@context, :update_submission_context_gradebook_url) %>" class="update_submission_grade_url">&nbsp;</a>
               <div class="turnitin_info_container"></div>
+              <div class="vericite_info_container"></div>
```
ktgeek: Since turnitin and vericite are mutually exclusive, let's use the same container for both.

app/views/jst/assignments/VeriCiteSettingsDialog.handlebars
```diff
@@ -0,0 +1,74 @@
+{{!
+  TODO: Replace this with new styelguide forms once the
+  upgrade to bootstrap is complete. The needed change would
+  be to add class="well form-horizontal bootstrap-form".
+}}
```
ktgeek: This comment needs to be removed since bootstrap is on its way out.

config/routes.rb
```diff
@@ -1875,6 +1878,7 @@
     post "caliper/:token", controller: :lti_api, action: :caliper_service, as: "lti_caliper"
     post "logout_service/:token", controller: :lti_api, action: :logout_service, as: "lti_logout_service"
     post "turnitin/outcomes_placement/:tool_id", controller: :lti_api, action: :turnitin_outcomes_placement, as: "lti_turnitin_outcomes_placement"
+    post "vericite/outcomes_placement/:tool_id", controller: :lti_api, action: :vericite_outcomes_placement, as: "lti_vericite_outcomes_placement"
```
ktgeek: This is for turnitin the LTI version, not the plugin.  This is unneeded for an non-LTI backend plugin like this.

lib/vericite.rb
```diff
@@ -0,0 +1,460 @@
+#
```
ktgeek: This file looks like a copy of the turnitin client with modifications to call the vericite gem code.  This probably should be redone to match the vericite way of operating.  The turnitin code is generally frowned upon, but was left as is under 1) it ain't broke and 2) it's going away.  Since #2 isn't true in this case, we'd like to see a better attempt at this file.

lib/vericite.rb
```diff
@@ -0,0 +1,460 @@
...
+      @endpoint = "/api.asp"
+      account_id = Canvas::Plugin.find(:vericite).settings[:account_id]
+      shared_secret = Canvas::Plugin.find(:vericite).settings[:shared_secret]
+      raise "Account ID required" unless account_id
+      raise "Shared secret required" unless shared_secret
+
+      @account_id = account_id
+      @shared_secret = shared_secret
+      @testing = testing
+      @functions = {
```
ktgeek: Example on why a rewrite is needed: this variable isn't used anywhere.

lib/vericite/response.rb
```diff
@@ -0,0 +1,112 @@
+#
```
ktgeek: This file needs to be revisited as well.  There are attr_accessors followed by get methods?  Some understanding of what is trying to be accomplished would be very helpful.

ktgeek:
Besides the comments left in the files, here's a few overall issues we'd like toes addressed:

- We'd like to see the gem packaged seperately and publicly and refer to that via the gemfile.  Would nessessitate a change to Gemfile.d/app.rb.  Also note: As a dependency, we didn't code review the gem.)

-  style conventions in new code should follow the ruby conventions.  For example, camel case variables in submission.check_verivite_status should be fixed.  (Not the only area, just one example.)

FYI, spec files weren't code reviewed deeply because the amount of implementation changes suggested would nessessitate changes to the specs.

app/models/submission.rb
ktgeek: I'm still a little confused by this.  The update code is acting in the context of a single submission, I missing where its cached for every submission on the assignment.

baholladay: Its a combination of "vericite_recheck_score", "check_vericite_status", and "vericite.generateReport" 

The walkthrough looks like:

1) code asks for vericite data like: submission.vericite_data
2) This calls the function vericite_data()
3) vericite_data determines whether this data was recently checked by calling "vericite_recheck_score". If so, it just returns the cached value
4) If not, it calls "check_vericite_status"
5) "check_vericite_status" has extra checks to make sure it only calls the VeriCite API every once in a while as well. If these checks say to call, then it calls vericite.generateReport
6) "vericite.generateReport" calls "sendRequest" where "command == :get_scores". In this function, it has another check for cached data. The requests calls VeriCite for all scores for that particular user. Then stores the results in a map based on each file. 

I will look into making step 6 even more generic and look up all scores for the entire course in one call. This would mean that at most, only one call per assignment can be made to VeriCite's API every 5 minutes, not matter how many users are loading that assignment.

baholladay:
Thanks for the code review! I will spend some time trying to go through the feedback and seeing what can be done.

app/coffeescripts/models/Assignment.coffee
baholladay: There is only one field required in the database for VeriCite:

alter table assignments ADD vericite_enabled boolean

"vericite_settings" is a logical function that uses the "turnitin_settings" database field. This allows VeriCite and Turnitin to have different advanced settings while still using the same database field.

app/coffeescripts/models/Assignment.coffee
baholladay: "vericite_settings" is a logical function that uses the "turnitin_settings" database field. This allows VeriCite and Turnitin to have different advanced settings while still using the same database field.

app/coffeescripts/views/assignments/EditView.coffee
baholladay: I'm struggling trying to figure out how to combine these two since I need VeriCite and Turnitin to have separate settings. This includes: "VeriCiteSettingsDialog" and "vericite_settings". Any suggestions?

app/coffeescripts/views/assignments/VeriCiteSettingsDialog.coffee
baholladay: I agree... this was too aggressively copied. I'll try to combine the two and at a minimum remove the unused TII code

app/models/submission.rb
cmatheson: SpeedGrader loads all the submissions for an assignment, so this will get called once for every student.  The Gradebook loads all submissions for all assignments, so this will get called many times for every student in the class.

My other concern is even if we do look up all scores in the course for each call, does that mean you have to load/update every submission in the course?  That is probably not going to work for large courses either.
.

baholladay: There should only be one call to the VeriCite API to get the latest scores for the entire page load and only if there is at least one VeriCite assignment. VeriCite scores are dynamic and can change based on edits within the report or if new material is discovered. All subsequent calls for each student will use the cached data from the first call.

baholladay: I just remembered that the "self.readonly" variable is set to "true" for the gradebook main page (i.e. the instructor view where all students and all assignments are listed). This means that this page will never consult VeriCite's API for the scores and only show what is stored in the Canvas DB (which it already has in the object). This is also the reason why I cache the scores based on SiteId + AssignmentId + UserID, since there is no other screen where VeriCite scores show up where it isn't just one user. Sorry about not remembering this until now, I wrote this code back in April and its hard to remember these small nuances. You can easily validate this by adding a log message in the beginning of this function:

Rails.logger.info "Readonly: #{self.readonly?}"

app/coffeescripts/views/assignments/VeriCiteSettingsDialog.coffee
baholladay: I deleted this file and made TurnitinSettingsDialog work with multiple plagiarism types

app/models/submission.rb
baholladay: This follows the same logic as mentioned above. It will do all of the work in bulk and store data in a cache. It will also only do API calls for individual pages and not the whole gradebook page

app/models/submission.rb
baholladay: the "vericite_data" function has a check t o make sure it only looks up API calls for the most current version:

"self.versions.current.number == self.version_number"

it also does all the lookups in bulk and stores the data in the cache

baholladay:
@ktgeek 
I have made the changes you requested in your code review. I will work on making the VeriCite gem packaged separately, but I figured that you could start going over the changes on the other code while I work on that packaging VeriCite as a gem.

baholladay:
@ktgeek @claydiffrient 
I have made all of the recommended changes and have completed making VeriCite a separately packaged GEM. Please continue with the review process!

claydiffrient:
@baholladay Would you mind squashing the 4 commits down to one?  It makes it much easier to run this through out internal build/review process (and keeps me from having to guess on conflicts 😄 )

baholladay:
@claydiffrient 
 I have squashed all commits into one. I originally included the commit history in case you wanted to see the revisions I made based on the code review.

app/controllers/submissions_controller.rb
```diff
@@ -473,23 +499,36 @@ def turnitin_report
...
+      else
+        # turnitin
+        @submission.resubmit_to_turnitin
+        messageKey = "resubmitted_to_turnitin"
+        messageDefault = "Successfully resubmitted to turnitin."
+      end
       respond_to do |format|
         format.html {
-          flash[:notice] = t('resubmitted_to_turnitin', "Successfully resubmitted to turnitin.")
+          flash[:notice] = t(messageKey, messageDefault)
```
jenseng: This doesn't work for I18n, since we do static analysis to extract the strings to be translated... `t()` calls always need a literal string. Additionally, specifying a key is not actually necessary anymore, so you could do something like:

```ruby
if type == 'vericite'
  @submission.resubmit_to_vericite
  message = t("Successfully resubmitted to VeriCite.")
else
  @submission.resubmit_to_turnitin
  message = t("Successfully resubmitted to turnitin.")
end
respond_to do |format|
  format.html {
    flash[:notice] = message
```

baholladay: @jenseng Thanks! That is good to know. I've made these changes and updated the PR.

app/models/submission.rb
cmatheson: I'm still not seeing how there is only one call in speedgrader.

app/views/plugins/_vericite_settings.html.erb
```diff
@@ -0,0 +1,59 @@
...
+      <td>
+        <%= f.text_area :pledge, :rows => '4' %>
+      </td>
+    </tr>
+    <tr>
+      <td><%= f.blabel :type, :en => "Students Can See the Originality Report" %></td>
+      <td><%= f.select :release_to_students, {"Immediately"=>'immediate', "After the assignment is graded"=>'after_grading', "After the Due Date"=>'after_due_date', "Never"=>'never'} %>
+    </tr>
+    <tr>
+      <td><%= f.blabel :host, :en => "Exclude Quotes" %></td>
```
cmatheson: these `blabel` calls are failing on our CI because they need to have unique labels (right now you are using `:host` several times.  Change this one to `:exclude_quotes` and fix any other duplicates as well.

app/models/submission.rb
```diff
@@ -482,6 +495,304 @@ def turnitinable_by_lti?
...
 
+  # VeriCite
+
+  # this function will check if the score needs to be updated and update/save the new score if so,
+  # otherwise, it just returns the vericite_data_hash
+  def vericite_data()
+    self.vericite_data_hash ||= {}
+    self.vericite_data_hash[:provider] = :vericite
+    # check to see if the score is stale, if so, fetch it again
+    updateScores = false
```
cmatheson: can we change any camelCase in this commit to snake_case ?

baholladay:
I have made the blabel and snake_case changes.

app/models/submission.rb
baholladay: It is already explicitly using the Submission's "!self.readonly?" attribute to choose whether it consults the database (readonly) or, if not read only, then it allows the tool to look up the scores from the VeriCite API. As a double safety measure, it only looks up the data if it is stale and if it hasn't just recently called the API. Also, when it does look up the scores, the service only looks up once for the entire class (all assignments and users in one call) and caches the results and then makes sure it doesn't call the API again for subsequent calls.

baholladay:
@ktgeek @cmatheson  I've made your requested changes. If you want me to make the VeriCite API call more explicitly stated, please review the attached patch on this comment. It adds a "lookup_api" parameter, which determines whether or not the "vericite_data" function will look up data in the VeriCite API or just rely on the Canvas DB. If you want this, I can include it on this PR. Please let me know.

[vericite_patch.txt](https://github.com/instructure/canvas-lms/files/407717/vericite_patch.txt)

ktgeek:
@baholladay Sorry for the delay in reply. With that last patch.txt file, we're about ready to accept it and move it into QA.  If would be helpful if you could make a new pull request with all the changes based on the latest master.

baholladay:
@ktgeek I've applied that last vericite_patch.txt file and rebased this PR. I also QA'ed one more time to make sure I didn't break anything in the rebase. This PR is ready for you and your team!

ktgeek:
Hey @baholladay!  During the move into QA we found a bunch of unit tests and selenium tests that weren't passing.  I've pulled out the relevant parts of the stack trace as well as the error that was recorded for you to address.

> spec/selenium/gradebook/gradebook2/gradebook2_turnitin_spec.rb:34
Selenium::WebDriver::Error::NoSuchElementError: Selenium::WebDriver::Error::NoSuchElementError

> /mnt/canvas/spec/selenium/assignments/assignments_turn_it_in_spec.rb:21:in `change_turnitin_settings'
/mnt/canvas/spec/selenium/assignments/assignments_turn_it_in_spec.rb:63:in `block (3 levels) in <top (required)>'
Selenium::WebDriver::Error::ElementNotVisibleError: Element is not currently visible and so may not be interacted with

> /mnt/canvas/spec/selenium/assignments/assignments_turn_it_in_spec.rb:21:in `change_turnitin_settings'
/mnt/canvas/spec/selenium/assignments/assignments_turn_it_in_spec.rb:80:in `block (2 levels) in <top (required)>'
Selenium::WebDriver::Error::ElementNotVisibleError: Element is not currently visible and so may not be interacted with

> spec/selenium/speedgrader_teacher_submission_spec.rb:226
Selenium::WebDriver::Error::NoSuchElementError: Unable to locate element: {"method":"css selector","selector":"#grade_container .submission_pending"}

> spec/selenium/speedgrader_teacher_submission_spec.rb:240
Selenium::WebDriver::Error::NoSuchElementError: Unable to locate element: {"method":"css selector","selector":"#grade_container .turnitin_similarity_score"}

> spec/selenium/speedgrader_teacher_submission_spec.rb:287
Selenium::WebDriver::Error::NoSuchElementError: Unable to locate element: {"method":"css selector","selector":"#grade_container .submission_error"}

> spec/selenium/speedgrader_teacher_submission_spec.rb:250
Selenium::WebDriver::Error::NoSuchElementError: Unable to locate element: {"method":"css selector","selector":"#grade_container .submission_error"}

> spec/apis/v1/submissions_api_spec.rb:1220
RSpec::Expectations::ExpectationNotMetError: expected #has_key?("turnitin_data") to return true, got false

I rebased against master before I confirmed the failing tests, so the pull request we're working on has a parent of 19bf7c867f57070b5ec81424452f129e75b82fc4

baholladay:
@ktgeek Is there a way we can get together virtually and talk about these errors? You can contact me by emailing support@vericite.com and we can try to set something up.

Is there documentation on how to run the selenium and unit tests? For example, I looked at the first error:

spec/selenium/gradebook/gradebook2/gradebook2_turnitin_spec.rb:34
Selenium::WebDriver::Error::NoSuchElementError: 
Selenium::WebDriver::Error::NoSuchElementError

That line is "icons = ffj('.gradebook-cell-turnitin')". However, looking at the attached screenshot, you can see that the plagiarism icons have that .gradebook-cell-turnitin class so I don't know why this would fail unless it just didn't work before or a previous test didn't work causing this part to not work.

screen shot 2016-08-24 at 10 39 50 am

baholladay:
@ktgeek I've updated one spec that may have caused the cascade of the other errors. Can you update and retry? I'll look into how to run these tests myself as well. Please contact me if you can so that we can be more efficient.

baholladay:
@ktgeek I've updated the PR to include the db migration script for VeriCite and the additional changes needed to make the selenium and unit tests to run for TII. Thanks for all of your help!

ktgeek:
Cut off one head of the hydra, two more grow in its place.  I'll be e-mailing you directly, @baholladay, looks like another pass is needed.

baholladay:
Thanks @ktgeek!

Keith,

I spent all this morning working on the unit tests and was able to figure out the issue. I have ran all of the following test successfully,

bundle _1.11.2_ exec rspec spec/selenium/client_apps/canvas_quizzes_spec.rb
bundle _1.11.2_ exec rspec spec/selenium/conversations/conversations_inbox_opt_out_spec.rb
bundle _1.11.2_ exec rspec spec/apis/v1/assignments_api_spec.rb
bundle _1.11.2_ exec rspec spec/selenium/quizzes/quizzes_log_auditing_spec.rb
bundle _1.11.2_ exec rspec spec/selenium/speedgrader_teacher_spec.rb
bundle _1.11.2_ exec rspec spec/apis/v1/stream_items_api_spec.rb
bundle _1.11.2_ exec rspec spec/apis/v1/assignments_api_spec.rb
bundle _1.11.2_ exec rspec spec/apis/v1/gradebook_history_spec.rb
bundle _1.11.2_ exec rspec spec/apis/v1/submissions_api_spec.rb
bundle _1.11.2_ exec rspec spec/lib/turnitin
bundle _1.11.2_ exec rspec spec/lib/turnitin_spec.rb
bundle _1.11.2_ exec rspec spec/selenium/assignments/assignments_turn_it_in_spec.rb
bundle _1.11.2_ exec rspec spec/selenium/gradebook/gradebook2/gradebook2_turnitin_spec.rb


However, there is one test that I do not believe is connected to my patch that keeps failing:

bundle _1.11.2_ exec rspec spec/selenium/differentiated_assignments/da_assignments_spec.rb

rspec ./spec/selenium/differentiated_assignments/da_assignments_spec.rb:125 # interaction with differentiated assignments Observer with student Assignment Show page and Submission page should allow previous submissions to be accessed on an inaccessible assignment
rspec ./spec/selenium/differentiated_assignments/da_assignments_spec.rb:51 # interaction with differentiated assignments Student Assignment Show page and Submission page should allow previous submissions to be accessed on an inaccessible assignment

I checked out a clean copy of Instructure's master and rebuilt/compiled everything and I still get those two errors. Can you see if you are able to reproduce this as well with both a clean master checkout (latest) as well as on my VeriCite plugin branch?

ktgeek:
Hey Bryan, it looks like the last of the unit test that wasn't passing was in our javascript side.  It looks like the test as written didn't match the code that was produced.  I'm guessing it was a cut and paste error.  I had the test match the code, and its now passed.  I've attached the tiny diff

[last.diff.zip](https://github.com/instructure/canvas-lms/files/457898/last.diff.zip)

baholladay:
@ktgeek Thanks! I've applied your patch to this PR. I believe all of the unit tests now pass! Thanks for your help on that.

ktgeek:
@baholladay, since we last talked in e-mail a day or two ago, have you pushed a new version of the pull request?

baholladay:
@ktgeek I made the changes to address the QA bugs and pushed them in this PR. I was not sure whether you wanted me to address the initial score concerns by adding an option to wait (or even if it was decided) and whether it would be part of this PR iteration or if it would be a 2nd PR iteration. I can meet with you if you want to discuss it.

ktgeek:
I don't think we need to meet.  I agree with doing the option to wait" as a second PR, this one is getting pretty big and going into an interactive period would be good.  We can hold off deploy on this side and, ideally, have the new PR built off this one.

Actually, maybe "option to not-wait" as I think we'd prefer waiting to be the default.

baholladay:
@ktgeek 
> I agree with doing the option to wait" as a second PR
Sounds good. We'll have to add a flag in our score API result, but its a simple change that won't break our existing LMS integrations and is a feature we could release to the other LMSs as well :)

> maybe "option to not-wait" as I think we'd prefer waiting to be the default
Making it the default sounds good. I can get this work started for the next PR.

Please have your QA team re-test the PR since I made those fixes to the bugs they found. It was the provider flag not being set in all cases, so make sure they test with a fresh database.

ktgeek:
I just repulled this into our local system and its not reporting any changes.  Can you double checked you pushed those up?

baholladay:
I double checked this PR code and see the changes I made. I also see the history of my commit has moved to between my comment 7 days ago and this one (cbf24d8), which GitHub does when you force update a commit. You can do an easy check by looking at line 11 in app/views/jst/assignments/VeriCiteSettingsDialog.handlebars I added "style="min-width: 18em;"" to address the concerns for the VC options dropdown not being wide enough:

<select id="settings_originality_report_visibility" style="min-width: 18em;

ktgeek:
Hmm... Okay, something is wonky here.  Looking at https://github.com/baholladay/canvas-lms/commits/vericite-master , which this pull request is against, the last update seems to be 9/9, which is the last time I pulled, and its identical to my pull from that day.  I'm not sure where/why I'm not seeing your update.

baholladay:
@ktgeek My last update was on Sept 8th which covered the bugs discovered during QA. I haven't started work on the "option to not-wait" since I thought that will be part of a different PR. What changes are you looking for?

ktgeek:
Oh sorry, I was confused.  For some reason I thought you had added more QA findings after 9/9.

baholladay:
@ktgeek Any update on the QA status? I have the code ready for the next PR which includes a configurable option to not show preliminary scores (default is to wait).

baholladay:
@ktgeek I have updated the PR based on the new round of QA feedback. It is ready for another round of QA. Thanks!

ccutrer:
@baholladay: I'm currently checking some Ruby 2.4 compatibility issues, and one of the things is that json must be updated to 2.0. Your vericite_api gem is currently requiring `(~> 1.4, >= 1.4.6)`. Would it be possible to update that to just `(>= 1.4.6)`?

baholladay:
I get a few warnings when I build the gem, but if you are ok with this, then I can proceed with updating the gem:

gem build vericite_api.gemspec 
WARNING:  description and summary are identical
WARNING:  open-ended dependency on json (>= 1.4.6) is not recommended
  if json is semantically versioned, use:
    add_runtime_dependency 'json', '~> 1.4', '>= 1.4.6'

ccutrer:
yup, that's what we're looking for. if you're really concerned about the warning, you could add '<3'. and of course, if your gem has any internal tests, they should pass with the json 2.0.2 gem

baholladay:
@ccutrer I've published an updated gem: vericite_api v1.2.0

Here is a patch to update it:

```
diff --git a/Gemfile.d/app.rb b/Gemfile.d/app.rb
index 5b9963d..8cccd9c 100644
--- a/Gemfile.d/app.rb
+++ b/Gemfile.d/app.rb
@@ -143,6 +143,6 @@ gem 'multipart', path: 'gems/multipart'
 gem 'paginated_collection', path: 'gems/paginated_collection'
 gem 'stringify_ids', path: 'gems/stringify_ids'
 gem 'twitter', path: 'gems/twitter'
-gem 'vericite_api', '1.1.0'
+gem 'vericite_api', '1.2.0'
 gem 'utf8_cleaner', path: 'gems/utf8_cleaner'
 gem 'workflow', path: 'gems/workflow'
```

PR #833, opened 2016-05-23
Fix yard-related errors/warnings

dgynn:
Yard is having trouble with multi-line statements that use dot-first notation and have comments at the end of line.
The commit moves to end-of-line dot notation (which is more standard in the codebase) and removes some obvious comments. Either change would resolve the issue.

This also switches one `@param` tag to `@argument` since it is not a param.

Test Plan:
  - Run `rake doc:api`
  - The Yard docs should be built without errors/warnings

### DIFF
--- a/app/controllers/sis_api_controller.rb
+++ b/app/controllers/sis_api_controller.rb
@@ -132,12 +132,12 @@ class SisApiController < ApplicationController
   end
 
   def published_assignments
-    Assignment.published
-      .where(post_to_sis: true)
-      .where(context_type: 'Course', context_id: published_course_ids)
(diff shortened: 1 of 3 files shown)


app/controllers/sis_api_controller.rb
```diff
@@ -133,11 +133,11 @@ def published_course_ids
 
   def published_assignments
     Assignment.published
```
ccutrer: you're missing the addition of a trailing dot on this line

dgynn: Argh. Sorry, pushed up now and tested with sis_api_spec.rb.

PR #875, opened 2016-08-19
Enable correct answer indicator in question banks

dgynn:
In the Question Banks, the correct questions are not marked when Question Details are shown. All of the data/JS is there, the container just needs another class added.

Test Plan:
  - Create a Question Bank in a Course and add a new Question
  - Click "Show Question Details"
  - Answers should now appear with indicator for correct answers

### DIFF
--- a/app/stylesheets/pages/quizzes/_quizzes.scss
+++ b/app/stylesheets/pages/quizzes/_quizzes.scss
@@ -532,7 +532,6 @@ li.quiz {
 .question_bank {
   border-bottom: 1px solid #b6babf;
   border-top: 1px solid #fff;
-  overflow: hidden;
   &:first-child {
     border-top: none;
   }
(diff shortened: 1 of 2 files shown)


claydiffrient:
@dgynn Seems like there's something causing the indicator to not work properly on Chrome.

app/stylesheets/pages/quizzes/_quizzes.scss
```diff
@@ -532,7 +532,6 @@ li.quiz {
 .question_bank {
   border-bottom: 1px solid #b6babf;
   border-top: 1px solid #fff;
-  overflow: hidden;
```
claydiffrient: The engineer that looked at it suspected this as the reason it's not working in Chrome.

dgynn:
I've double checked this and I don't see what is wrong. One thing I have seen is that if you click "Show Question Details" before the page has fully loaded the answer arrows will not be displayed although the answers will be. Toggling the checkbox off and back on will display the arrows. And so will waiting for the page to fully load before clicking the box.

Is it possible that it was tested too fast? And did it work in a different browser?

claydiffrient:
@dgynn I'm told that it worked it Firefox without any problems.  I've been informed that product would like to make a decision on this since it changes behavior.  I'll let you know when a decision is made.

PR #877, opened 2016-08-19
Prevent warning when redefining YAML

dgynn:
Test Plan:
  - Run `rails c`
  - The console should load without "warning: already initialized constant YAML"
  - All tests should still pass

### DIFF
--- a/lib/canvas_yaml.rb
+++ b/lib/canvas_yaml.rb
@@ -28,6 +28,7 @@ require 'syck' # so we can undo all the things before something else requires it
 if defined?(YAML::ENGINE)
   YAML::ENGINE.yamler = 'psych'
 else
+  Object.send(:remove_const, :YAML)
   YAML = Psych # :/
 end
 


lib/canvas_yaml.rb
```diff
@@ -28,6 +28,7 @@
 if defined?(YAML::ENGINE)
```
ccutrer: while you're in the area - can you confirm that this case will never happen again (I'm pretty sure it was only for Ruby 1.9, which Canvas no longer allows), and remove this dead code

dgynn: Ah, yes, and in looking into that, I found that there is one other thing that needs to be added here.
`Object#to_yaml` also needs to be re-patched. I'll drop the dead code and add the to_yaml fix.

dgynn: My mistake. I see to_yaml is fixed below. Commit is updated with dead code removed.

dgynn:
It turns out that code is needed for Ruby 2.1. I've restored the original version of the commit.

PR #878, opened 2016-08-19
Sort courses by id desc so the most recent course is first

dgynn:
The courses on the account page are listed without any specified ordering. 

A case could be made to sort alphabetically or by last updated date or by published state. I would like to suggest that sorting by ID descending (or created_at) would be the most helpful and would make it easy to find newly created courses.

Test Plan:
  - From the accounts page, create a new course or two
  - Reload the page
  - The newly created course(s) should be listed first

### DIFF
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -481,7 +481,7 @@ class Account < ActiveRecord::Base
 
   def fast_course_base(opts)
     columns = "courses.id, courses.name, courses.workflow_state, courses.course_code, courses.sis_source_id, courses.enrollment_term_id"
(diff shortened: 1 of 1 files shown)


claydiffrient:
@dgynn In discussing this with our product team, they would prefer that sorting default to alphabetical, and that we allow sorting by the various other information types if the user elects to do so.

There has been some work on a new course/user search page for the account page (check it out with the Account Course and User Search feature flag in a development environment).  That is likely where we'd want to handle this.  I'm going to go ahead and close this PR for now.  Thanks for your contribution!

PR #882, opened 2016-08-24
Live events - Attachments

simong:
test plan:
  * create attachments by uploading course files
  * update display name of course files
  * delete some course files
  * delete a course wiki page
  * ensure appropriate messages are in the kinesis queue


simonista:
Thanks @simong, I'm going to run this through our full test suite, and get a couple other engineer's eyes on it.  I'll let you know if any issues come up.

lib/canvas/live_events.rb
```diff
@@ -101,6 +101,27 @@ def self.get_submission_data(submission)
...
+      unlock_at: attachment.unlock_at,
+      lock_at: attachment.lock_at,
+      updated_at: attachment.updated_at,
+      hidden: attachment.hidden
+    }
+  end
+
+  def self.assignment_created(assignment)
+    post_event_stringified('assignment_created', get_assignment_data(assignment))
+  end
```
simonista: looks like this method got duplicated from up above, can you remove it?

lib/canvas/live_events.rb
```diff
@@ -184,6 +205,23 @@ def self.wiki_page_deleted(page)
     })
   end
 
+  def self.attachment_created(attachment)
+    post_event_stringified('attachment_created', get_attachment_data(attachment))
+  end
+
+  def self.attachment_updated(attachment, old_display_name)
+    payload = get_attachment_data(attachment)
+    if old_display_name
```
simonista: Do you have a use case for needing the old_display_name, or were you just following the pattern of updated events?  One of the reasons this isn't public yet is that we've been debating whether or not to include these old_* type fields in final public api, and just asking people to track what changed themselves if they care about that.  I'd love to hear more about your use case here if you have one and how onerous it would be for you if it wasn't there.

simonista:
Looks like `spec/models/content_migration/course_copy_spec.rb:97` failed with `Failure/Error: expect(att.reload).to be_hidden, got nil`.  Interesting interaction, looks like because we now call `.hidden` when an attachment is created, it caches https://github.com/instructure/canvas-lms/blob/stable/app/models/attachment.rb#L1089, and so it's already set as nil when the test checks for hidden?  It'd be easy to just fix the test, but let me think a little more about whether caching that value that early might be a problem in other places.

lib/canvas/live_events.rb
simong: I was just following the pattern for updated events. We don't really have a need for knowing what the old display name was, we just need to be notified that an attachment has a new one.

I'm not all that familiar with Ruby or Rails, but would it make sense to provide the `changes` hash for `update` events? That way you're not cherry picking which attributes are getting carried over. Again, I'm not familiar with whether there's sensitive data on any of the models that have an `update` event though.

lib/canvas/live_events.rb
```diff
@@ -101,6 +101,23 @@ def self.get_submission_data(submission)
...
   end
 
+  def self.get_attachment_data(attachment)
+    {
+      attachment_id: attachment.global_id,
+      user_id: attachment.global_user_id,
+      display_name: LiveEvents.truncate(attachment.display_name),
+      filename: LiveEvents.truncate(attachment.filename),
+      context_type: attachment.context_type,
+      context_id: attachment.context_id,
```
simong: Jut an FYI here: I added `context_type` and `context_id` on the attachment data as we need it on our end, but I'm not sure whether that's necessary? From re-reading the docs again, it looks like they might be part of the body metadata which should be available across all types of events?

I currently don't have access to an SQS queue where these messages should be getting delivered so it's hard to tell what might be required.

lib/canvas/live_events.rb
```diff
@@ -101,6 +101,23 @@ def self.get_submission_data(submission)
...
+      content_type: attachment.content_type,
+      folder_id: attachment.folder_id,
```
simong: On a slightly tangential note, we're currently only interested in attachment events that take place on attachments in courses that are exposed in the files tool. Meaning:
 - files that get uploaded in the files tool
 - files that get uploaded as a discussion, assignment, .. attachment and end up at `unfiled` (or somewhere exposed in the files tool).

Is it safe to assume that an attachment with `context_type = 'Course'` always gets exposed in the course files tool? AFAICT, submissions are recorded with a `User` context, but I might not get exposed to all the places that have attachments logic.

Our current plan is to perform an extra API request to `/api/v1/folders/:folderId` and double check the name of the folder starts with `course files`. It would be nice not to have to do this.

lib/canvas/live_events.rb
zwily: The `context_type` and `context_id` included in the `metadata` body is only there if the event was triggered by a web request, and will be the context for the request itself, which is likely not what you want. Including these here appear to be what you want.

lib/canvas/live_events.rb
zwily: I think you're correct about submissions being on `User` context, but I don't know for sure that `'Course'` files are always exposed in the course files tool. I *am* pretty sure that anything with `context_type = 'Course'` and a non-null `folder_id` will be part of the course files though.

simong: Cool, that should simplify a few things (and minimise the number of API requests we'll have to make)

lib/canvas/live_events.rb
```diff
@@ -101,6 +101,23 @@ def self.get_submission_data(submission)
...
+      unlock_at: attachment.unlock_at,
+      lock_at: attachment.lock_at,
+      updated_at: attachment.updated_at,
+      hidden: attachment.hidden
```
zwily: For now, we should probably leave `hidden` off. The problem is that accessing `hidden` will load the entire folder parentage if it's not already loaded. That is likely not a problem (I suspect it's usually loaded), but we're trying to be super careful about only sending data in live events that we're sure is already loaded. Plus, this is breaking a spec. If you really need this attribute, we can dig in and make sure we're comfortable adding it, but the easiest way forward would be to just remove it.

lib/canvas/live_events.rb
zwily: We do want to cherry-pick at this point. We're starting conservatively. Down the road, we may want to look at just sending the `changes` hash.

zwily:
This is now passing all tests. Now it's up for another pass of code review and QA.

zwily:
@simong - I don't think we have a CLA on file for you. You'll need to fill one out (see https://github.com/instructure/canvas-lms/wiki/FAQ#id-like-to-contribute-to-canvas) and send it to opensource@instructure.com before we can merge. Thanks!

simong:
@zwily I've submitted the contributors agreement to `opensource@instructure.com`. It should be a company wide one.

simong:
I've merged up master into this branch.

I don't want to abuse this PR as support but it looks like the `wiki_page_update` events currently don't truncate the title or body. Is that by design and if so, can we rely on always getting the full page body?

simonista:
Merged to master in af60de97bb4777e04a3340faa23068c932fe13f7

PR #901, opened 2016-09-28
For students show the display name in the roster

eriko:
This also applies to other non privileged users.  In all other areas of canvas that are accessible to students the users display name is shown and not the legal name.  For students who name in use is not their legal name this allows them to be contactable inside canvas by their name. This also should apply to groups. The display name is sent in the JSON as short_name and is easily available.  For teachers and other privileged users the legal name will be displayed.  That said a good option would be to display the (display name) in parens after the link.

I looked through the selenium tests and people_spec was the closest to this that I could find but it was only testing in the context of people as a teacher or ta. So I am not sure where to put tests.

### DIFF
--- a/app/views/jst/courses/roster/rosterUser.handlebars
+++ b/app/views/jst/courses/roster/rosterUser.handlebars
@@ -2,7 +2,11 @@
   {{>avatar}}
 </td>
 <td>
+  {{#if canViewSisId}}
   <a href="{{html_url}}" class="roster_user_name">{{name}}</a>
+  {{else}}
+  <a href="{{html_url}}" class="roster_user_name">{{short_name}}</a>
+  {{/if}}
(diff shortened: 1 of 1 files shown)


eriko:
This PR would resolve this. 
https://community.canvaslms.com/ideas/3806

tierra:
Hi Erik, thanks for your interest in addressing this. Can you please rebase your changes on top of the master branch?

Additionally, it's probably worth noting that this could potentially (and optimally) be tested using view specs. I'm sure I've seen a bunch that render these views as an authenticated student vs teacher.

I've confirmed that we already have your CLA on file, thanks.

eriko:
Ok I will do that. 

Can you point out were the view spec's are.  I tried to find those and failed.  I found a couple things that tested other parts of the page but not the list.

eriko:
redone as https://github.com/instructure/canvas-lms/pull/914

PR #914, opened 2016-10-07
For students show the display name in the roster redo

eriko:
I botched the rebase so here is #901 done on master.
https://github.com/instructure/canvas-lms/pull/901

repeated from the other PR
This also applies to other non privileged users. In all other areas of canvas that are accessible to students the users display name is shown and not the legal name. For students who name in use is not their legal name this allows them to be contactable inside canvas by their name. This also should apply to groups. The display name is sent in the JSON as short_name and is easily available. For teachers and other privileged users the legal name will be displayed. That said a good option would be to display the (display name) in parens after the link. (I have changed to do this in the new PR)

I looked through the selenium tests and people_spec was the closest to this that I could find but it was only testing in the context of people as a teacher or ta. So I am not sure where to put tests.

### DIFF
--- a/app/views/jst/courses/roster/rosterUser.handlebars
+++ b/app/views/jst/courses/roster/rosterUser.handlebars
@@ -2,7 +2,11 @@
   {{>avatar}}
 </td>
 <td>
-  <a href="{{html_url}}" class="roster_user_name">{{name}}</a>
+  {{#if canViewSisId}}
+  <a href="{{html_url}}" class="roster_user_name">{{name}}</a> ({{short_name}})
+  {{else}}
+  <a href="{{html_url}}" class="roster_user_name">{{short_name}}</a>
+  {{/if}}
   {{#if isPending}}<span class="label label-info" title="{{#t "pending_acceptance_explanation"}}This user has not yet accepted the invitation to the course{{/t}}">{{#t "pending_acceptance_of_invitation"}}pending{{/t}}</span>{{/if}}
   {{#if isInactive}}<span class="label" title="{{#t}}This user is currently not able to access the course{{/t}}">{{#t}}inactive{{/t}}</span>{{/if}}
 </td>


tierra:
I've pushed these changes up for code review.

It would still definitely help push this through with spec coverage though, I'm not sure this will be accepted without new specs. Maybe ember/handlebars makes this difficult to test with a view spec, so maybe it should be covered with a new student-authenticated selenium test. The teacher-authenticated context in the `people_spec.rb` selenium suite should at least provide a decent starting point. If you're new to running selenium specs, I highly recommend running them using docker: https://github.com/instructure/canvas-lms/blob/stable/doc/development_with_docker.md#selenium

eriko:
I will take a shot at it.  It looks like I will have to dig back pretty far as it looks like neither name ever gets tested and may not be set.

app/views/jst/courses/roster/rosterUser.handlebars
```diff
@@ -2,7 +2,11 @@
   {{>avatar}}
 </td>
 <td>
-  <a href="{{html_url}}" class="roster_user_name">{{name}}</a>
+  {{#if canViewSisId}}
+  <a href="{{html_url}}" class="roster_user_name">{{name}}</a> ({{short_name}})
```
tierra: This should handle leaving the parenthesis off when short_name is blank.

app/views/jst/courses/roster/rosterUser.handlebars
```diff
@@ -2,7 +2,11 @@
...
+  {{else}}
+  <a href="{{html_url}}" class="roster_user_name">{{short_name}}</a>
```
tierra: This should still fall back to `name` if `short_name` is blank.

app/views/jst/courses/roster/rosterUser.handlebars
tierra: Also, in the interest of internationalization, this should now use something similar to this as well (when short_name isn't blank):
```
{{#t}}<a href="{{html_url}}" class="roster_user_name">{{name}}</a> ({{short_name}}){{/t}}
```

eriko:
Ok I have spent most of the day trying to get things set up build some tests and the existing tests are failing.  Dory, built in to dinghy is not functioning to resolve host and at this point I need to get back to working on my own rails apps.  

I hope this or a similar change makes it in to the codebase as we can not give students access to the people page in a course as it does not list students working name.

app/views/jst/courses/roster/rosterUser.handlebars
eriko: I am not sure how display_name (shortname) can ever be blank but I will cover that.

app/views/jst/courses/roster/rosterUser.handlebars
eriko: What actual needs to happen here is that if name and short_name are the same then the (shortname) should not be shown. There is no equality comparison in handbars so I am not sure how to catch that with out changing what was is sent in the api.

eriko:
Ok I have updated my branch (patch-8) with an updated squashed and rebased change. https://github.com/eriko/canvas-lms/tree/eriko/patch-8  It does not seem to be updating the pull request as I had expected it to.

app/views/jst/courses/roster/rosterUser.handlebars
tierra: Yeah, maybe this could really use a new `roster_user_name` display variable that's already built in the API.

eriko: If `roster_user_name` was left nil with `name` and `short_name` were the same it would probably make the logic a bit easier also.

eriko:
replaced with #937

PR #930, opened 2016-10-18
Fixing bug with loading a VeriCite assignment as a student after disa…

baholladay:
…bling the VeriCite plugin

### DIFF
--- a/app/views/assignments/_vericite.html.erb
+++ b/app/views/assignments/_vericite.html.erb
@@ -5,7 +5,7 @@
     </td>
   </tr>
 <%- end %>
-<%- if !@context.vericite_pledge.empty? %>
+<%- if @context.vericite_pledge.present? %>
   <tr>
     <td colspan="2">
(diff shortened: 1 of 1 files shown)


tierra:
@baholladay Thanks for following up on fixes to your previously contributed code! Your commit message doesn't meet the standards outlined in our [CONTRIBUTING.md](https://github.com/instructure/canvas-lms/blob/stable/CONTRIBUTING.md#submitting-pull-requests).  If you'd update this, we can get this pulled in and get the code review process started.

Considering this is a bugfix, it might indicate that we really need spec coverage of some functionality that clearly wasn't covered in the initial contribution, but probably should have been. It doesn't look like we have view specs for the similar Turnitin view, but it is covered by a couple selenium tests. At a minimum, we should probably consider similar selenium specs for this VeriCite view, but ideally it would be better if this was covered by view specs.

baholladay:
@tierra I have updated the commit message for this PR

PR #937, opened 2016-11-01
Display the short name (display name) for a user. (replacement for #914)

eriko:
For students they will see the display name.
For people that can see sis data they will see Name (Display Name).

Display name is only shown if it is different than Name

Tests for the view both as a teacher and a student.


eriko:
So here is the main issue with how this works.  The `roster_name` field in the json signals two things. The first thing is signals is that if it is not there that either the `short_name` field is equal to `name` or that `short_name` is empty.  This lets the handlebars use the `name` and turn off the display of the `short_name` for teachers on the roster (People) page.  The second thing it signals is by its presence in the feed that the `short_name` is different than the `name`. Now I could just always send the `roster_name` with it defaulting to `name` if `short_name` was empty but this does not let me optionally show the `short_name` on the teachers version of the roster (People) page so that they can call the person by their name and not their other name.

This is latter is important for people that are changing their name but it has not made it through the bureaucracy yet.

eriko:
So any thoughts on this?

ccutrer:
While the overall goal is sound, I'm not a fan of this approach (adding an additional field to API responses that's identical to one of the other fields). Instead, I'd rather the front end showed full name if it's available, otherwise short name.

As to why students should see the short name, not the full name - is that because students shouldn't be allowed to see the full name? Or simply that they would normally prefer the short name. If the former, then full name should not be returned in the API result at all for those not allowed to see it. If it's a preference thing, that should be handled in the front end, similar to how you already are, just reversing the logic of which name is preferred.

Finally, being able to view sis data is not a good differentiator of teacher vs. students. Even though that's the default, it's not always the case that teachers can see SIS data, or students can't see it. You need to get down the basic question of "what is required to see full name (by default?). Is there already a specific permission that matches this question?"

eriko:
As to why:
In all other areas of canvas that are accessible to students the users display (short) name is shown and not the legal name. For students who use a name that is not their legal name this allows them to be contactable inside canvas by their name on the People tab. This also should apply to groups.   In our case we have a large number of trans students and while we are federally legally required to use their legal name (dead name) in the internal systems we also track and use their preferred name.  In the case of canvas we feed this as the short name.  Currently we have to leave the People's tab off as it does not show the students name and instead shows what the student refers  to as their dead name .

As to the implementation:
I am not fully happy with it also (see my comment on Nov 1, 2016) It feels too side effect based to be a good answer.  You are correct that using the availability of the sis data is not a good way to choose but it was there and a non invasive way to decide. I was trying to change as little as possible to get the effect.  At the time I was asked to do the testing in selenium and so that is where I did it. 

As to the comments on the code:
Yes I think those are pretty much spot on.

What I can do now:
Wrote this up mainly to show the inconsistency with the current interface and give a rough draft of how to fix it.  In the months since I did this I have moved machines and removed the vm's that I was doing development/testing for it.  I have unloaded from my head all the context to do the work. It will take me quite some time to reload all that work into my head.  We are in the process of ramping up a full Identity management project and I do not know how much time I can give to this as it has been 4 months since I last looked at canvases codebase.  

I would not offended at all if someone who actually works on the codebase ditched this and did it in a cleaner way. I think it would probably be a better choice. If no one will do this let me know and I will try to find some time in the next couple months.

omarkhan:
Thanks @eriko. I don't know that we'll have time to work on this ourselves, but I'll leave it open for now in case you want to revisit it later.

ccutrer:
A trans student is probably the best use-case for the difference between full name and display name that I've seen yet :)

But I still haven't answered the question for myself if normal students should not be able to see full name (permissions issue, probably due to FERPA or some other regulation, not allowed to expose it at all in the API) or they simply _prefer_ not to (the API can expose it, but the UI defaults to not showing it). I'm not necessarily looking for an answer from you; it can come from Instructure's product team as well. Of course, if it's a legal issue that's applicable to our customers, it's much easier to get Instructure to prioritize working on this. If we can't make that case, then it becomes a nice-to-have, and may sit for a long time.

eriko:
@ccutrer my thought is that only people that legally need to see the legal name should see the legal name.  This is already true elsewhere in canvas.  

@omarkhan I will see what I can get done.

simonista:
I'm going to close this since there's been no update since April, but feel free to reopen if you do end up having time to get back to it.  Thanks again for the PR.

eriko:
We have decided to just use the students preferred name and not their legal name in canvas. Closing this is fine.  That said you should consider doing some thing like this anyways.

PR #960, opened 2016-12-07
Fix external error reporting via HTTP POST

johnnaegle:
When enabling and testing the external error reporting plugin, I found that
it had a typo: `Net:HHTP` instead of `Net::HTTP`.

Test Plan:
  - Enable external errpor rerpoting to a URL
  - Send an error report with `ErrorReport.last.send_to_external`
  - Expect the configured URL to receive the POST

### DIFF
--- a/app/models/error_report.rb
+++ b/app/models/error_report.rb
@@ -228,7 +228,7 @@ error_id: #{error_report.id}
       params = {}
       params[config[:subject_param]] = error_report.subject
       params[config[:body_param]] = body
-      Net::HHTP.post_form(URI.parse(config[:url]), params)
+      Net::HTTP.post_form(URI.parse(config[:url]), params)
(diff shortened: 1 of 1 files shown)


ccutrer:
This is ancient code. If you configure /plugins/canvas_ticketing_by_web_post and /plugins/ticketing_system to use the Web Post connector, that should be satisfy the functionality you're looking for, no?

johnnaegle:
Yes, one of those other plugins should satisfy what I want to do.  I think I take Web Post through AWS Lambda and get it to a slack channel.  

However, the syntax error is still fixed in my commit.

ccutrer:
I'm going to close this PR then. I plan on ripping out this code. I'm not aware of anyone using it, and the ticketing system stuff is an adequate replacement.

PR #961, opened 2016-12-09
Validate submission types for VeriCite assignments

baholladay:
VeriCite only supports file submissions and text entry submission types. When creating or editing an assignment, validate the submission types are only of these type.

Test Plan:
1) Enable VeriCite in the plugins page. You can use dummy data since we are only validating the saving of the assignment
2) Test saving a combination of "Enable VeriCite Submissions" with the available submission types. If VeriCite is enabled, only "Text Entry" and or "File Uploads" (both are ok) are allowed. Including any other submission type should warn the user and not allow them to save the Assignment.
3) Test saving an assignment without VeriCite enabled (both with and without the VeriCite plugin enabled). Make sure you can save any combination of submission types.


ktgeek:
hey Bryan, I get this one and the other one started through the process.

baholladay:
Keith! As always, let me know if you have any questions.

PR #962, opened 2016-12-09
Show VeriCite reports for Peer Review assessments

baholladay:
When an assignment has both VeriCite and Peer Review enabled, allow reviewers to see the VeriCite reports.

Test Plan:
1) Enable VeriCite in the account plugins page
2) Create an assignment with both VeriCite and Peer Review enabled
3) Submit as multiple students
4) Assign peer reviews (manually or automaticaly)
5) As a student, view a peer review and make sure the VeriCite report is available
6) Modify the assignment's VeriCite advanced settings and set Students Can See the Originality Report to "none"
7) Make sure the students can not see the VeriCite reports in the Peer Review page

### DIFF
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -326,13 +326,15 @@ class Submission < ActiveRecord::Base
       plagData = self.vericite_data_hash
       @submit_to_vericite = false
       settings = assignment.vericite_settings
+      type_can_peer_review = true
     else
       plagData = self.turnitin_data
       @submit_to_turnitin = false
       settings = assignment.turnitin_settings
+      type_can_peer_review = false
     end
     return plagData &&
-    user_can_read_grade?(user, session) &&
+    (user_can_read_grade?(user, session) || (type_can_peer_review && user_can_peer_review_plagiarism?(user))) &&
     (assignment.context.grants_right?(user, session, :manage_grades) ||
       case settings[:originality_report_visibility]
        when 'immediate' then true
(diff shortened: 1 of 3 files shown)


ktgeek:
Pulled in for review

app/models/submission.rb
```diff
@@ -344,6 +346,16 @@ def can_view_plagiarism_report(type, user, session)
     )
   end
 
+  def user_can_peer_review?(user)
+    allowed = false
+    if assignment.peer_reviews && self.assigned_assessments
+      self.assigned_assessments.each do |review|
+        allowed = allowed || user.id == review.user_id
```
ktgeek: I'm not sure I'm following what you're trying to do here.

If there are multiple assigned assessments, you'll just keep reassigning allowed.  Perhaps you want the .any? method here instead of .each?

Also, it looks like you really want allowed ||= user.id == review.user_id.  But again, I think the .any? method is more of what you're looking for.

app/models/submission.rb
```diff
@@ -344,6 +346,16 @@ def can_view_plagiarism_report(type, user, session)
     )
   end
 
+  def user_can_peer_review?(user)
```
ktgeek: This is a helper method just for the permission above, correct?  Let's change the name of this to be user_can_peer_review_plagiarism or something similar.

app/models/submission.rb
```diff
@@ -326,13 +326,15 @@ def can_view_plagiarism_report(type, user, session)
       plagData = self.vericite_data_hash
       @submit_to_vericite = false
       settings = assignment.vericite_settings
+      enabled_for_peer_review = true
```
ktgeek: Maybe a better name is type_can_peer_review for this variable.

app/models/submission.rb
baholladay: >Perhaps you want the .any? method here instead of .each?
Yup! I was just writing it out long hand, didn't realize Ruby has a shorthand for it... thanks!

baholladay:
@ktgeek I've updated the PR with your suggestions

baholladay:
I'll see if I can get a unit test included

ktgeek:
QA kicked this back with: "Students reviewing a submission don't see any vericite icons."

baholladay:
@ktgeek As I was working on the unit test, I discovered why the QA found what they did. I didn't have a large enough test data set for the peer review. I've updated the logic in "user_can_peer_review_plagiarism" and everything should be working now. Also, I've added a selenium unit test.

PR #963, opened 2016-12-14
Bigbluebutton

jfederico:
This PR includes a number of improvements to the integration with BigBlueButton through the  plugin although it also includes some additions to the conference module itself.

As discussed with Chris Ward, from the UI perspective the enhancements can be seen in three shot videos we already shared with him (we can send you guys the links through slack).


app/controllers/plugins_controller.rb
```diff
@@ -40,6 +40,7 @@ def update
     if find_plugin_setting
       @plugin_setting.disabled = value_to_boolean(params[:plugin_setting][:disabled]) if params[:plugin_setting] && !params[:plugin_setting][:disabled].nil?
       @plugin_setting.posted_settings = params[:settings] || {} unless @plugin_setting.disabled
+      @plugin_setting.settings[:secret] ||= params[:settings][:secret] if @plugin_setting.name == "big_blue_button" && @plugin_setting.settings
```
ccutrer: You should not be special casing individual plugins in this controller. You should be modifying the validator for that plugin instead

app/models/big_blue_button_conference.rb
```diff
@@ -21,15 +21,21 @@
...
-  after_destroy :end_meeting
-  after_destroy :delete_all_recordings
+  before_destroy :end_meeting
+  before_destroy :delete_all_recordings
+
+  RECORDING_OPTIONS = {
+    1 => 'Show record option',
+    2 => 'Hide record option (true by default)',
+    3 => 'Record everything'
+  }
```
ccutrer: these are not translated

claydiffrient:
I would suggest that some of these commits be squashed down.  Our general guidelines in our [CONTRIBUTING.md](https://github.com/instructure/canvas-lms/blob/stable/CONTRIBUTING.md) suggest:

> Your pull request should generally consist of a single commit. This helps keep the git history clean by keeping each commit focused on a single purpose. If you have multiple commits that keep that focus then that is acceptable, however "train of thought" commits should not be in the history.

This PR is definitely large enough to merit multiple commits however some definitely seem to be "train of thought" type commits, for example "Fixed merge conflicts", "Fixed typo", "Merged upstream", "Css fixes", and so forth.

So if you can go through and clean that up a little, that would be much appreciated :).

jfederico:
Hey @claydiffrient  @ccutrer, we got it!

As you can see now we have done forced push into the same branch with only two commits. All @erickadbay 's history now belongs to the past :)

So, you can check it out from here or if you think is better, you can close this PR and we create a new one. Whatever you think is best.

claydiffrient:
@jfederico Generally pull requests are only accepted to the `master` branch rather than `stable` I did the Github change base thing, which looks like it might have brought up some conflicts.  Might be best to rebase everything on your end on the `master` branch and re-push clearing up any potential conflicts.

jfederico:
Hi @claydiffrient I just created a new PR to master. It has some minor changes for fixing issues with the settings not being saved with the version in master.

You may want to close this one and keep in thouch through the new PR:

Improvements to BigBlueButton plugin #976

PR #968, opened 2017-01-10
Import and export VeriCite advanced settings

baholladay:
When importing an assignment into a course, the VeriCite advanced settings were not imported

Test Plan:
1) Enable VeriCite in the plugins page. You can use dummy data since we are only validating the saving of the assignment
2) Create two courses and add an instructor to both
3) In course A, create an assignment with submission type=online and check "Enable VeriCite Submissions".
4) Click "Advanced VeriCite Settings" and make changes to the default settings and click "Update Settings".
5) Save assignment
6) In course B, import the assignments from Course A (https://guides.instructure.com/m/4152/l/57077-how-do-i-copy-content-from-another-canvas-course)
7) In course B, verify that the assignment was created and that the VeriCite settings are correct, including the advanced settings

### DIFF
--- a/spec/models/content_migration/course_copy_assignments_spec.rb
+++ b/spec/models/content_migration/course_copy_assignments_spec.rb
@@ -153,6 +153,12 @@ describe ContentMigration do
       assignment_model(:course => @copy_from, :points_possible => 40, :submission_types => 'file_upload', :grading_type => 'points')
       @assignment.turnitin_enabled = true
       @assignment.vericite_enabled = true
+      @assignment.vericite_settings = {
+          :originality_report_visibility => "after_grading",
+          :exclude_quoted => '1',
+          :exclude_self_plag => '0',
+          :store_in_index => '1'
+      }
       @assignment.peer_reviews = true
       @assignment.peer_review_count = 2
       @assignment.automatic_peer_reviews = true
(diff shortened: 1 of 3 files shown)


spec/models/content_migration/course_copy_assignments_spec.rb
```diff
@@ -175,7 +181,11 @@
 
       new_assignment = @copy_to.assignments.where(migration_id: mig_id(@assignment)).first
       attrs.each do |attr|
-        expect(@assignment[attr]).to eq new_assignment[attr]
+        if @assignment[attr].class == Hash
```
ktgeek: I don't think this block of code is needed.  Where you getting failure states without doing this?  The eq spec is smart enough to match hashes pretty well.

spec/models/content_migration/course_copy_assignments_spec.rb
```diff
@@ -153,6 +153,12 @@
       assignment_model(:course => @copy_from, :points_possible => 40, :submission_types => 'file_upload', :grading_type => 'points')
       @assignment.turnitin_enabled = true
       @assignment.vericite_enabled = true
+      @assignment.vericite_settings = {
```
ktgeek: We probably should change this into two specs.  As a whole we treat turnitin and vericite enabled to be mutually exclusive.  That said, since they share a common backend this is probably okay, but it does go against our general behavior.

That said, this code probably has a lifetime of about a year, so I'm inclined to be okay with it.

spec/models/content_migration/course_copy_assignments_spec.rb
baholladay: I was running into an issue with "eq" not working because sometimes the hash had String keys and other times it was ":" keys... this was the only way I could get it to work. Did you not have any issues?

baholladay: its similiar to why you would use "hash.with_indifferent_access"

baholladay:
I've rebased

PR #972, opened 2017-01-19
VeriCite: resubmit option should be available for any incomplete reports

baholladay:
The resubmit option doesn't show up unless the submissions process runs successfully. This isn't helpful for instructors who have submissions that the process never successfully kickekd off and results in orphaned submissions that never get submitted to VeriCite. This patch makes sure that all the correct meta data is saved to the submission before the process even begins, so that the instructor will always see the resubmit option if the process errored out for any reason.

Test Plan:

1) Enable VeriCite in the plugins page. You can use dummy data since we are only validating the submission process
2) Create an assignment
3) Submit to this assignment as a student (one for attachments and one for inline submissions)
4) As the instructor, you should immediately see a gray pending icon in both the gradebook overview and the speedgrader page. In the speedgrader page, you should be able to click the pending icon and see an option to resubmit to VeriCite.

### DIFF
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -781,6 +781,16 @@ class Submission < ActiveRecord::Base
   VERICITE_JOB_OPTS = { :n_strand => 'vericite', :priority => Delayed::LOW_PRIORITY, :max_attempts => 2 }
 
   def submit_to_vericite_later
+    # make sure any missing statuses get set to "pending"
+    vericite_assets.each do |a|
+      asset_data = self.vericite_data_hash[a.asset_string] || {}
+      #we only want to change the status if it is not present or a hard reset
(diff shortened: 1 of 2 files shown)


ktgeek:
I think the need for this one has been lessoned by fixing the underlying issue that was causing this to come up.  Check out a4217dba2b9a3a16371cfded35997e9720999c6f

baholladay:
I agree and this technically should never be needed if everything is working right. I'm ok with not committing this PR, it was a suggestion which would make the "resubmit" option appear always and immediately no matter what, even if everything was broken or messed up.

omarkhan:
Given that the underlying issue has been fixed, I will close this unless anyone objects?

PR #974, opened 2017-01-23
Conformed code style to eslint rules

amilajack:


claydiffrient:
@amilajack Thanks for the Pull Request!  Unfortunately, we aren't going to accept it at this time.  

We haven't fully settled on our standard set of eslint rules to enforce at the moment, and we see linting as an incremental improvement rather than something that should be done all at once.  I'm going to update our contribution guidelines to reflect this for future contributions.  Thanks again!

PR #991, opened 2017-02-13
VeriCite report links do not work when a user clicks to view the report

baholladay:
This seems to be a regression from https://github.com/instructure/canvas-lms/commit/e087da776f2a8037886510ab1677773244ee5653 where it deletes the provider identifier for the originality_data, which is used to determine whether the link is a VeriCite or TII link

Test Plan:
1) enable VeriCite plugin
2) create an assignment with VeriCite enabled
3) submit to the assignment and wait for report to be ready
4) click the link to view the report
5) make sure report works and doesn't give an error saying report is missing
(also check anything relate to TII and viewing reports)

### DIFF
--- a/app/models/submission.rb
+++ b/app/models/submission.rb
@@ -561,7 +561,6 @@ class Submission < ActiveRecord::Base
       }
     end
     ret_val = turnitin_data.merge(data)
-    ret_val.delete(:provider)
     ret_val
   end
 


baholladay:
This test case is an invalid test case:

spec/models/submission_spec.rb:
it "filters out :provider key and value" do

The provider key should never be filtered. This key distinguishes the plagiarism provider (e.g. VeriCite or TII). Another words, it should only be one of two values but always set.

westonkd:
@baholladay Thank you for bringing this issue to our attention, we appreciate your contribution! Could you share the place in Canvas you clicked the link to view the report and ran into the issue? I’m assuming it was from the modal in gradebook, but wanted to double check. Could you also post what the href of the link was? Thank you!

westonkd:
Thank you for bringing the issue to our attention! I was able to reproduce the issue from the `/courses/:id/assignments/:id/submissions/:id` page in Canvas. 

We are currently working on a feature that requires filtering the provider when the `submission#originality_report` method is called, but it looks like we can approach fixing the problem you reported from a different direction. 

Were there any other locations in Canvas had this issue? I was not able to identify the problem from other pages but wanted to double check with you.

baholladay:
@westonkd 
Thanks for the quick response!

> Were there any other locations in Canvas had this issue?
I have checked and that seems to be the only page where the URLs are trying to load TII instead of VeriCite. The logic to display the VeriCite URL needs to know the provider to determine whether it is a TII or VeriCite URL in: show_preview.html.erb (search for VeriCite or TII) as well as the report URL lookup logic that is triggered from pages like this in submissions_controller.rb:legacy_plagiarism_report (see the type parameter is supposed to be the value of the provider). I'm not sure if my PR is the correct fix (it fixes the VeriCite issue, but I can't test TII), especially since you say you are working on a new feature that requires filtering.

westonkd:
Great, thanks for confirming that is the spot you are seeing the issue. I've created a fix by modifying some of the view logic that solves the issue for Vericite, TII, and leaves us the filtering we need in the method. Thanks!

baholladay:
@westonkd That's great! I guess this PR can be closed. Do you know when all of these changes go into our Canvas dev instance? We would like to test all of our workflows as soon as its available.

westonkd:
New code is placed in your beta environment every three weeks (on Monday) as indicated by the Beta Release Notes which can be found here: http://releases.canvaslms.com. Thanks!

simonista:
I _believe_ 72af3ef3d06096cf1800eefdd4180c543169a27e is the fix for this.

PR #992, opened 2017-02-13
Use mail_gate to restrict the delivery of mail to only whitelisted email

johnnaegle:
> MailGate is an additional delivery method for the Mail gem that lets you restrict the delivery of mail to only whitelisted emails. Ideal for staging environments where you may be using production data and do not want them to receive emails from your mailers when you submit comments, contact forms, or anything else that may trigger mail delivery.

This PR adds mail_gate as a supported delivery method specifically so staging environments only send email to whitelisted email addresses.

Details:
- Adds a new delivery_method to outgoing_mail.yml: mail_gate
- When mail_gate is the delivery method, and the email is whitelisted, use SMTP as the delivery mechanism
- outgoing_mail.yml.example has settings for configuring mail_gate.  Additional options can be found with the mail_gate gem

Test Plan:
- Configure outgoing email with mail gate and a whitelisted email domain
- Send messages to a whitelisted domain and another domain.  Expect to only receive email on the whitelisted domain

### DIFF
--- a/Gemfile.d/mail_gate.rb
+++ b/Gemfile.d/mail_gate.rb
@@ -1,0 +1,1 @@
+gem 'mail_gate'
--- a/config/initializers/outgoing_mail.rb
+++ b/config/initializers/outgoing_mail.rb
@@ -7,7 +7,7 @@ require 'net/smtp'
 config = {
   :domain => "unknowndomain.example.com",
   :delivery_method => :smtp,
-}.merge((ConfigFile.load("outgoing_mail") || {}).symbolize_keys)
+}.merge((ConfigFile.load("outgoing_mail") || {}).deep_symbolize_keys)
 
 [:authentication, :delivery_method].each do |key|
   config[key] = config[key].to_sym if config.has_key?(key)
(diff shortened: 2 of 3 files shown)


omarkhan:
Thanks for the contribution @johnnaegle, and sorry for the delay in responding. I don't know that this is something we can incorporate and maintain going forward, given the lack of tests and the fact that similar functionality can be achieved in other ways (by adding a filter at the mail server level, for example). I will close the pull request for now, but if you feel strongly about this feel free to reopen it and I'll see if I can get any buy-in from product.

tierra:
I'm not sure if you've noticed this, but the local docker compose development environment is pre-configured to use MailCatcher if you enable [this override](https://github.com/instructure/canvas-lms/blob/stable/docker-compose/mailcatcher.override.yml).

johnnaegle:
it looks like outgoing_mail.yml uses mailgate by default for staging:

```ruby
# Use mail_gate to block emails to any non @instructure.com email address.
# for whitelisted emails, smtp is used with the delivery_settings.  These
staging:
  delivery_method: mail_gate
  whitelist: /\@instructure.com/
  subject_prefix: '[Staging] '
  delivery_settings:
    address: "smtp.example.com"
    port: 25
    user_name: "user"
    password: "password"
    authentication: "plain" # plain, login, or cram_md5
    domain: "example.com"
    outgoing_address: "canvas@example.com"
    default_name: "Instructure Canvas"
```

PR #1004, opened 2017-03-15
VeriCite: add student preview advanced option

baholladay:
If an assignment has VeriCite's student preview enabled, all VeriCite reports for this assignment will be hidden from the instructor until the student submits the draft or the due date has passed. Scores and links will still be visible.

Test Plan:

1) Enable VeriCite in plugins. Use a real key and secret and make sure the "show student preview" option is checked.
2) Create an assignment with VeriCite enabled and student preview enabled. Make sure due date is set and the advanced VeriCite settings is set to show reports immediately for students (not required, just suggested)
3) Make sure VeriCite assignment settings save and can be edited
4) When a student submits to a student preview enabled assignment, they should see an indicator stating student preview is enabled before submitting
5) When a paper is submitted and a score is ready, if the assignment has student preview enabled, the scores should show a "draft" icon next to the score everywhere for students and instructors. Only caveat is the instructor gradebook overview page (since this page doesn't look up scores directly from VeriCite but relies on the stale Canvas database scores). Also, make sure that an assignment without student preview set works and doesn't show the "draft" icon next to the score.
6) A student can either submit the draft as a final submission (via the VeriCite report UI) or you can wait for the due date to pass. In either case, afterwards, the "draft" icon should disappear. There is a 20 minute cache on report scores/draft status, so you may have to wait 20+ minutes afterwards.
7) make sure other plagiarism service integrations work and assignments/clients w/o plagiarism services enabled work.


ktgeek:
This is in process of being reviewed by our product folks.

ktgeek:
Based on the coming of the new plagiarism framework, we're passing on this for now.

PR #1006, opened 2017-03-19
VeriCite: add encryption flag for submissions to specify encryption type

baholladay:
Test Plan:
1) Enable VeriCite in plugins. Use a real key and secret.
2) Create an assignment with VeriCite enabled
3) Submit to this assignment
4) Make sure report is generated (~10-15 mins)


lib/vericite.rb
```diff
@@ -263,7 +263,7 @@ def sendRequest(command, args)
           external_content_data.upload_content_length = args[:psize]
           report_meta_data.external_content_data = external_content_data
           # @return [Array<ExternalContentUploadInfo>]
-          data, status_code, headers = vericite_client.reports_submit_request_context_id_assignment_id_user_id_post(context_id, assignment_id, user_id, consumer, consumer_secret, report_meta_data)
+          data, status_code, headers = vericite_client.reports_submit_request_context_id_assignment_id_user_id_post(context_id, assignment_id, user_id, consumer, consumer_secret, report_meta_data, {:'encrypted' => true})
```
ktgeek: `{ encrypted: true }` should be the format for this hash.

Also, if that is going to be hard coded, is this something the vericite gem should be handling?  Is there ever a time where you wouldn't want it encrypted?

lib/vericite.rb
```diff
@@ -273,7 +273,7 @@ def sendRequest(command, args)
           end
           data.each do |externalContentUploadInfo|
             #API will return an upload URL to store the submission (throws an exception if it fails)
-            res = api_client.uploadfile(externalContentUploadInfo.url_post, args[:pdata])
+            res = api_client.uploadfile(externalContentUploadInfo.url_post, args[:pdata], externalContentUploadInfo.headers)
```
ktgeek: Our linter noticed that res is an unused variable, and I concur.  Should this be assigned to the response from above?  If its truly not used (and i realize the existing code was like that) we should probably just ignore the return value.

lib/vericite.rb
baholladay: I'll update and test the format. 

Yes, it is always encrypted. This is essentially a flag to indicate to our api to use the new call/response headers for encryption security. We are phasing out the old method and we use this flag to be able to accomidate the hundreds of LMS integrations updating at different times.  I could tuck this away within the vericite gem if you prefer.

baholladay:
@ktgeek I've updated the PR with your recommendations

lib/vericite.rb
ktgeek: Sorry for not responding sooner.  If that's going to be the new normal, throwing that into the gem makes the most since, since we've got it set in the Gemfile we won't have to worry about any backwards compatibility issues (within canvas.)

PR #1007, opened 2017-03-21
VeriCite: allow institutions to enable VeriCite at the subaccount level

baholladay:
Currently, institutions can only enable VeriCite at the top level in a "all or nothing" fashion. This feature will allow institutions to enable VeriCite at a subaccount level. There are many use cases for this including: each sub account is a different school and either 1) not all schools have subscribed to VeriCite 2) each school wants a unique key to VeriCite. Also, there are K-12 schools who do not need VeriCite enabled for the lower grades.

Test Plan:

1) Repeat the following steps with multiple plugin settings:
a) plugin setting 1: Set "Global VeriCite Account ID" and "Global VeriCite Shared Secret". Leave "Account VeriCite Credentials" empty
b) plugin setting 2: Set "Global VeriCite Account ID" and "Global VeriCite Shared Secret". Set "Account VeriCite Credentials".
c) plugin setting 3: Leave "Global VeriCite Account ID" and "Global VeriCite Shared Secret" empty. Set "Account VeriCite Credentials".
d) plugin setting 4: Leave "Global VeriCite Account ID" and "Global VeriCite Shared Secret" and "Account VeriCite Credentials" empty

2) Create a new assignment. Make sure that the "Enable VeriCite Submissions" shows or is hidden based on account settings. (caution, if using the same user, it could be automatically "checked" since this setting is saved as a user preference)
3) If you can, create the VeriCite assignment
4) Submit both an attachment and inline submission
5) Make sure the reports show up (could be 10+ minutes depending on server load, to speed this up, you can also set "Show Preliminary Score" to true in the plugin settings)

### DIFF
--- a/app/models/assignment.rb
+++ b/app/models/assignment.rb
@@ -450,10 +450,10 @@ class Assignment < ActiveRecord::Base
   end
 
   def create_in_vericite
-    return false unless Canvas::Plugin.find(:vericite).try(:enabled?)
+    return false unless VeriCite.enabled_for_account?(self.context.account_id)
     return true if self.turnitin_settings[:current] && self.turnitin_settings[:vericite]
     vericite = VeriCite::Client.new()
-    res = vericite.createOrUpdateAssignment(self, self.turnitin_settings)
+    res = vericite.createOrUpdateAssignment(self, self.turnitin_settings, self.context.account_id)
 
     # make sure the defaults get serialized
     self.turnitin_settings = turnitin_settings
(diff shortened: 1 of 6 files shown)


omarkhan:
Thanks for the contribution @baholladay! I'll run this by our product people. It won't get merged without tests, but let me check if its something we want to move forward with first.

ktgeek:
Due to the API transition being eminent, we're going to decline this one.

PR #1008, opened 2017-03-21
Add notes field and usage stats to Developer Keys

grahamb:
This pull request enhances the Developer Keys page in two ways:

## Notes field

A notes field is added on the `developer_keys` table. When creating a key, there is sometimes a need to record more information (for example, a ticket number for the request, information on how the key is to be used, etc). The notes field is exposed in the developer key listing, in the API response, and in the dialog for creating/editing a key.

## Usage stats

While adding the notes field, I found a `TODO` marker for "add some better stats here, like token count, last usage, etc.". Mark that `TODO` as `DONE`! This PR adds two methods to the `DeveloperKey` model: `access_token_count` and `last_used_at`. These fields are returned in the API response, and are displayed in the Stats column on the Developer Key page.

---
screenshot 2017-03-21 16 55 19

screenshot 2017-03-21 16 55 28

----

I have left my commits un-squashed in case there are changes required. I can squash them into one commit prior to merging.

A signed CLA is on file under my employer, Simon Fraser University.

### DIFF
--- a/app/models/developer_key.rb
+++ b/app/models/developer_key.rb
@@ -106,6 +106,10 @@ class DeveloperKey < ActiveRecord::Base
     account.try(:name)
   end
 
+  def last_used_at
+    self.access_tokens.maximum(:last_used_at)
+  end
+
   class << self
     def find_cached(id)
       global_id = Shard.global_id_for(id)
--- a/app/views/jst/developer_key.handlebars
+++ b/app/views/jst/developer_key.handlebars
@@ -28,23 +28,33 @@
     <td>
         <div class="details">
             <div>
-               {{#t "identifier"}}ID: {{id}}{{/t}}
+               {{#t}}ID: {{id}}{{/t}}
             </div>
             <div>
-               {{#t "api_key"}}Key: <span class='api_key'>{{api_key}}</span>{{/t}}
+               {{#t}}Key: <span class='api_key'>{{api_key}}</span>{{/t}}
             </div>
             <div>
                 {{#if redirect_uri}}
-                    {{#t "redirect_uri"}}URI: {{redirect_uri}}{{/t}}
+                    {{#t}}URI: {{redirect_uri}}{{/t}}
                 {{/if}}
             </div>
         </div>
     </td>
     <td>
         <div>
-            {{#t "created"}}Created:{{/t}} {{datetimeFormatted created_at}}
+          {{#t}}Access Token Count: {{access_token_count}}{{/t}}
         </div>
-        <!-- TODO: Add some better stats here, like token count, last usage, etc. -->
+        <div>
+            {{#t}}Created:{{/t}} {{datetimeFormatted created_at}}
+        </div>
+        <div>
+            {{#t}}Last Used:{{/t}} {{#ifExists last_used_at}}{{last_used_at}}{{else}}Never{{/ifExists}}
+        </div>
+    </td>
+    <td class='notes'>
+      <div>
+        {{#t}}{{notes}}{{/t}}
+      </div>
     </td>
     <td class='links'>
         <a href="#" class="edit_link" aria-label="{{#t}}Edit key {{name}}{{/t}}" title="{{#t "edit_key"}}Edit this key{{/t}}"><i class="icon-edit standalone-icon"></i></a>
@@ -53,6 +63,6 @@
         {{else}}
           <a href="#" role="checkbox" aria-checked="true" aria-label="{{#t}}Activate key {{name}}{{/t}}" class="deactivate_link" title="{{#t}}Deactivate this key{{/t}}"><i class="icon-lock standalone-icon"></i></a>
         {{/if}}
-        <a href="#" class="delete_link" aria-label="{{#t}}Delete key {{name}}{{/t}}" title="{{#t "delete_key"}}Delete this key{{/t}}"><i class="icon-trash standalone-icon"></i></a>
+        <a href="#" class="delete_link" title="{{#t}}Delete this key{{/t}}"><i class="icon-trash standalone-icon"></i></a>
     </td>
 </tr>
--- a/app/controllers/developer_keys_controller.rb
+++ b/app/controllers/developer_keys_controller.rb
@@ -88,6 +88,6 @@ class DeveloperKeysController < ApplicationController
   end
 
   def developer_key_params
-    params.require(:developer_key).permit(:api_key, :name, :icon_url, :redirect_uri, :redirect_uris, :email, :auto_expire_tokens)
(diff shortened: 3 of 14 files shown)


app/views/jst/developer_key.handlebars
```diff
@@ -42,9 +42,19 @@
     </td>
     <td>
         <div>
+          {{#t "access_token_count"}}Access Token Count: {{access_token_count}}{{/t}}
```
claydiffrient: So here and in other places, I notice that you are using explicit keys for the i18n stuff.  I'm not sure if we ever published anything externally for that, but explicit keys are deprecated in favor of some really nice automatic stuff.

So we'd prefer this be something like `{{#t}}Access Token Count: {{access_token_count}}{{/t}}` instead.

app/models/developer_key.rb
```diff
@@ -101,6 +101,14 @@ def account_name
     account.try(:name)
   end
 
+  def access_token_count
+    self.access_tokens.count
```
tpickett66: Count can be quite sad when using postgres, this is a perfect case for using a [counter_cache](http://guides.rubyonrails.org/association_basics.html#options-for-belongs-to-counter-cache)

app/models/developer_key.rb
```diff
@@ -101,6 +101,14 @@ def account_name
...
+  end
+
+  def last_used_at
+    self.access_tokens.order(last_used_at: :desc).first&.last_used_at
```
tpickett66: Since we're ordering by `last_used` we'll need an index on it, also this can be made to not load the full ActiveRecord Object:

```ruby
def last_used
  self.access_tokens
    .order(last_used_at: :desc)
    .limit(1)
    .pluck(:last_used_at)
    .first
end
```

With an index Postgres will most likely just scan the index and pull the value from there w/o looking at the full table files when done this way.

app/views/jst/developer_key.handlebars
grahamb: Yeah, I was just going with the style that was already being used in the file.  I can see the new way documented in https://github.com/instructure/canvas-lms/wiki/I18n#handlebars.

I'll change my additions to remove the explicit keys. Do you want me to change the existing ones while I'm at it?

claydiffrient: Oh okay cool.  I'm glad we have some external facing docs around that 😅.

It's such a toss up for me.  Part of me says kill the existing keys so we are closer to having none, but that does add a bit more to the scope of your change.  Part of me says not to worry about it and if those strings ever change then we'll kill it.  I don't _think_ we'll reject the PR either way.

grahamb: I get keeping the scope small; I briefly considered re-writing the entire thing in React and ditching the handlebars/coffeescript, but I figured I'd start small ;) On the other hand, it's a pretty minor change, all considered.

app/models/developer_key.rb
jenseng: Just a couple clarifications:Postgres likely wouldn't do a table scan since it already has the index

jenseng: Since there's no index on developer_key_id, the performance may not be better with a last_used_at index alone (and the planner might be right in not using it), since you'll have interleaved tuples from all developer keys. I'd recommend adding a composite index on (developer_key_id, last_used_at). This (index scan) will offer better read performance than two indexes (bitmap index scan, where it has synthesize two giant bitmaps), which in turn is generally better than a sequential scan.

ccutrer: even better:
```ruby
access_tokens.maximum(:last_used_at)
```

tpickett66:
I knew you'd come in with something.

app/views/jst/developer_key.handlebars
grahamb: Fixed in 893745e

app/models/developer_key.rb
grahamb: I've got this implemented. It looks like it I need to call `reset_counters` on each `DeveloperKey` in order to get the initial counts populated for the existing keys. Would this be an appropriate use for a `DataFixup`-type migration?

grahamb: Implemented using `.maximum` in c1d8bcf.

app/models/developer_key.rb
grahamb: Implemented using .maximum in c1d8bcf.

app/models/developer_key.rb
grahamb: Implemented in b71c2f8.

grahamb: @tpickett66 Actually, not quite. When checking the test I wrote for `access_token_count`, I found that when a new token is created, the `developer_keys.access_token_account` field is `NULL`, not `0`. 

I could fix this in a couple of ways:
- setting the default value for `developer_keys.access_token_count` to `0` and making it `NOT NULL`; or
- by calling `DeveloperKey.reset_countners(self.id, :access_tokens)` (probably in an `after_create` method).

I'm leaning towards the default value unless you've got a different suggestion.

tpickett66:
The updates look good, just need to add the index to `access_tokens.{developer_key_id, last_used_at}` and we can get this merged.

tpickett66:
As it turns out I wasn't up to date on the internal discussion on `counter_cache`, I just got up to date there and there may be some issues using it here. We need to do some investigation before we proceed.

grahamb:
OK, standing by…

tpickett66:
Ok, after discussing this we've decided to let this go forward with one additional change: we really need an index on `access_tokens` spanning `developer_key_id` and `last_used_at`.

grahamb:
No problem. I'll have a commit with that shortly.

grahamb:
@tpickett66 two questions:
* predeploy or postdeploy migration?
* disable transactions for the migration?

tpickett66:
Make it a `predeploy`, you can just use a `change` method and you probably don't need to supply the name unless you're getting an error from the automatic name being too long.

```ruby
def change
  add_index :access_tokens, [:developer_key_id, :last_used_at], algorithm: :concurrently
end
```

grahamb:
I edited it to use `change` while you were looking at it; I'll push a commit using `predeploy` in a sec.

grahamb:
Index implemented in b4bc2e7.

grahamb:
@tpickett66 do you need anything else for this? I can squash everything down to one commit if you want.

claydiffrient:
@grahamb Go ahead and do the squash/rebase and I'll get this merged in for you :)

grahamb:
@claydiffrient done!

grahamb:
@claydiffrient just checking to see where this is in the pipeline

claydiffrient:
@grahamb Hmm... looks like it got stale.  I don't think anyone from QA picked it up.  Probably my bad, looks like I missed getting an internal ticket made for it (so they probably didn't see it).  I'll get it  going again.

PR #1018, opened 2017-04-05
Ignore query parameters for developer key redirects

StefSchenkelaars:
When using the new developer key `redirect_uris`, the final redirect uri should exactly match one of the provided uris. However, some LTI apps require request specific parameters to allow cookieless sessions. This results in request specific oauth redirect uris (or at least the query parameters). This is a known issue and therefore this is even discussed in [the api documentation](https://canvas.instructure.com/doc/api/file.oauth.html#oauth2-flow-2) which states:

> If your application passed a state parameter in step 1, it will be returned here in step 2 so that your app can tie the request and response together.

This is true for the old legacy `redirect_uri` but it does not hold for the new `redirect_uris` method. This is because the full redirect uri is checked to the `redirect_uris` as can be seen [here](https://github.com/instructure/canvas-lms/blob/400389b0ca7a1ab8a0c2f1ca6857cd85250cf579/app/models/developer_key.rb#L141). This should not look at any query parameters (in my opinion) but certainly not look at the state query parameters since this is described in the api documentation.

This simple fix just removes the query parameters before checking if it is a valid uri. The input is converted to a string before splitting to prevent an error on nil as an input. I've tested this with an LTI app with cookieless sessions and thus request specific query parameters in the oauth redirect url.

Then a side note:
- The rubocop linter fails with `Perceived complexity for redirect_domain_matches? is too high. [8/7]`. This however also fails when I reject my own changes. This linter failure can be fixed by moving the legacy url check to a seperate method but I am not sure if this is desired.

### DIFF
--- a/app/models/developer_key.rb
+++ b/app/models/developer_key.rb
@@ -145,7 +145,7 @@ class DeveloperKey < ActiveRecord::Base
   # verify that the given uri has the same domain as this key's
   # redirect_uri domain.
   def redirect_domain_matches?(redirect_uri)
-    return true if redirect_uris.include?(redirect_uri)
+    return true if redirect_uris.include?(redirect_uri.to_s.split('?').first)
 
     # legacy deprecated
     self_domain = URI.parse(self.redirect_uri).host
--- a/spec/models/developer_key_spec.rb
+++ b/spec/models/developer_key_spec.rb
@@ -75,13 +75,13 @@ describe DeveloperKey do
 
     it "does not allow subdomains when it matches in redirect_uris" do
       key = DeveloperKey.create!(redirect_uris: "http://example.com/a/b")
-      expect(key.redirect_domain_matches?("http://example.com/a/b")).to eq true
+      expect(key.redirect_domain_matches?("http://example.com/a/b")).to be_truthy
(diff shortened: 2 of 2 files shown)


ccutrer:
I'll be happy to update documentation to make this clearer (but as you mentioned, the docs already mention the state parameter), but the OAuth spec is pretty clear on this. Redirect URIs must match exactly. If you need to pass additional information through, use the state parameter. You can encode whatever and however much information you want into this single parameter. This definitely matches how Google works, so saying that a 3rd party tool needs to rely on additional parameters is not accurate.

StefSchenkelaars:
@ccutrer The docs showed this example for the authorize url:

> GET https://<canvas-install-url>/login/oauth2/auth?client_id=XXX&response_type=code&redirect_uri=https://example.com/oauth_complete&state=YYY

Since the `&state=YYY` was behind the `redirect_uri` parameter, I assumed it should be part of the `redirect_url` parameter. So maybe the docs can be simple improved by changing the above to:

> GET https://<canvas-install-url>/login/oauth2/auth?client_id=XXX&response_type=code&state=YYY&redirect_uri=https://example.com/oauth_complete

PR #1023, opened 2017-04-13
Allow fullscreen for YouTube links

johndyer:
YouTube videos embedded on Pages are locked at 425x344. This allows the
fullscreen button to function so you don’t have to click thorough to
youtube.


omarkhan:
Thanks for this @johndyer! Let me run this by a product person to see if there is any reason we would *not* want to do this.

omarkhan:
Got the go-ahead for this. I'm shepherding it through our review process now. Thanks again for the contribution!

omarkhan:
@johndyer this is now in master. It isn't showing up as a merge as it went through our internal review system and got cherry-picked straight to master, but you should still have full credit for the commit. Thanks for the contribution!

PR #1034, opened 2017-05-08
VeriCite: Renaming plugin settings from "account id" to "consumer key"

baholladay:
VeriCite: Renaming plugin settings from "account id" to "consumer key" to be consistent

Test Plan:

1) Load the VeriCite plugins page: /plugins/vericite
2) Verify that the first text field is named "VeriCite Consumer Key" instead of "VeriCite Account Id"

### DIFF
--- a/app/views/plugins/_vericite_settings.html.erb
+++ b/app/views/plugins/_vericite_settings.html.erb
@@ -42,7 +42,7 @@
       </td>
     </tr>
     <tr>
-      <td><%= f.blabel :account_id, :en => "VeriCite Account ID" %></td>
+      <td><%= f.blabel :account_id, :en => "VeriCite Consumer Key" %></td>
       <td>
         <%= f.text_field :account_id %>
       </td>


config/locales/en-AU.yml
```diff
@@ -12903,7 +12903,7 @@ en-AU:
       default_assignment_settings: "**Default Assignment Settings**"
       default_comments_settings: "*Text entered here will be displayed to students when submitting for an assignment.*"
       labels: 
-        account_id: "VeriCite Account ID"
+        account_id: "VeriCite Consumer Key"
```
ktgeek: These translation files are generated elsewhere and don't need to receive hand updates.

config/locales/en-GB.yml
```diff
@@ -12899,7 +12899,7 @@ en-GB:
       default_assignment_settings: "**Default Assignment Settings**"
       default_comments_settings: "*Text entered here will be displayed to students when submitting an assignment*"
       labels: 
-        account_id: "VeriCite Account ID"
+        account_id: "VeriCite Consumer Key"
```
ktgeek: These translation files are generated elsewhere and don't need to receive hand updates.

baholladay:
@ktgeek updated

ktgeek:
Running it through internal tests now

PR #1040, opened 2017-05-16
Fix for the OAuth2 provider's remember authorization checkbox label

jackhughesweb:
This commit fixes the display of the OAuth 2 provider

The checkbox label for remembering an authorization displayed a hash
containing the en value and CSS class, and should instead be displaying
just the text contained in the en value.

 Test Plan:
   - Login to an external service, using Canvas as an OAuth2 provider,
and using a non-empty scope
   - The text next to the checkbox should now display “Remember my
authorization for this service”, and not the hash “{:en=>"Remember my
authorization for this service", :class=>"checkbox”}”


simonista:
Hey @jackhughesweb, thanks so much for the PR. 👍 from me. I'll run it through our test suite and make sure everything passes.

PR #1041, opened 2017-05-17
Handle a developer key having no notes

grahamb:
This fixes a bug introduced in #1008. If a developer key had no notes (e.g. the column was NULL), the template would render `[missing %{notes} value]`. This fixes that behaviour.

screenshot 2017-05-17 13 30 07

Test plan:
- Have a canvas instance with existing developer keys with NULL values in the notes column
- Observe that the notes column is empty
- Add a note for a key
- Observe that the notes for that key are displayed

### DIFF
--- a/app/views/jst/developer_key.handlebars
+++ b/app/views/jst/developer_key.handlebars
@@ -48,13 +48,15 @@
             {{#t}}Created:{{/t}} {{datetimeFormatted created_at}}
         </div>
         <div>
-            {{#t}}Last Used:{{/t}} {{#ifExists last_used_at}}{{last_used_at}}{{else}}Never{{/ifExists}}
+            {{#t}}Last Used:{{/t}} {{#ifExists last_used_at}}{{last_used_at}}{{else}}{{#t}}Never{{/t}}{{/ifExists}}
         </div>
     </td>
     <td class='notes'>
-      <div>
-        {{#t}}{{notes}}{{/t}}
-      </div>
+      {{#if notes}}
+        <div>
+            {{notes}}
+        </div>
+      {{/if}}
     </td>
     <td class='links'>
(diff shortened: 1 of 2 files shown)


simonista:
Thanks @grahamb, do you think you could add an automated test for this?

grahamb:
Sure. It'll have to be a selenium test since it has to render the handlebars template, correct?

simonista:
You can render a handlebars template in a javascript test

grahamb:
@simonista I've added a few tests; one for a NULL in the notes field, one for an empty-but-not-NULL notes field, and one for a key with a note. Let me know if anything needs to change.

app/views/jst/developer_key.handlebars
```diff
@@ -52,9 +52,11 @@
         </div>
     </td>
     <td class='notes'>
-      <div>
-        {{#t}}{{notes}}{{/t}}
-      </div>
+      {{#if notes}}
+        <div>
+            {{#t}}{{notes}}{{/t}}
```
simonista: This doesn't need to be wrapped in a translate block, because there is no actual content to translate (no label or anything).

app/views/jst/developer_key.handlebars
```diff
@@ -52,9 +52,11 @@
         </div>
```
simonista: "Never" (on the line above this, can't comment there for some reason) needs to be `{{#t}}Never{{/t}}`

app/views/jst/developer_key.handlebars
grahamb: Fixed in a2d2307

app/views/jst/developer_key.handlebars
grahamb: Fixed in a2d2307

simonista:
hey @grahamb, we're ready to merge this, would you mind squashing and rebasing?

grahamb:
@simonista Done!

simonista:
Merged in 7f2031f79642959670de936f75a33ea03bad246e

PR #1044, opened 2017-05-19
VeriCite: get all scores for an assignment in bulk

baholladay:
When loading speedgrader in a class with a lot of students, the VeriCite API is called per student. This patch gets all student scores in one call for the assignment and stores them in the cache for quick retrieval

Test Plan:
1) Setup an instance with VeriCite's native integration (plugins)
2) Setup a course with hundreds of students
3) Create an assignment with VeriCite enabled
4) Submit to this assignment for each student
5) wait for reports to be ready
6) load speedgrader as instructor. This page load time shouldn't be any (much) slower than if the assignment didn't have VeriCIte enabled.

### DIFF
--- a/lib/vericite.rb
+++ b/lib/vericite.rb
@@ -281,13 +281,18 @@ module VeriCite
           context_id = course.id
           assignment_id = assignment.id
           user_id = user.id
-          user_score_cache_key = "vericite_scores/#{consumer}/#{context_id}/#{user_id}"
-          # first check if the cache already has the score:
-          user_score_map = Rails.cache.read(user_score_cache_key)
-          if user_score_map == nil
+          user_score_cache_key_prefix =  "vericite_scores/#{consumer}/#{context_id}/#{assignment_id}/"
+          users_score_map = {}
+          # first check if the cache already has the user's score and if we haven't looked up this assignment lately:
+          users_score_map["#{user_id}"] = Rails.cache.read("#{user_score_cache_key_prefix}#{user_id}")
+          if users_score_map["#{user_id}"].nil? && Rails.cache.read(user_score_cache_key_prefix) == nil
+            #we already looked up this user in Redis, don't bother again (by setting {})
+            users_score_map["#{user_id}"] ||= {}
             # we need to look up the user scores in VeriCite for this course
             # @return [Array<ReportScoreReponse>]
-            data, status_code, headers = vericite_client.reports_scores_context_id_get(context_id, consumer, consumer_secret, {:'user_id' => user_id})
+            data, status_code, _headers = vericite_client.reports_scores_context_id_get(context_id, consumer, consumer_secret, {:assignment_id => assignment_id})
+            #keep track of the assignment lookup api call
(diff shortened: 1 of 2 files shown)


lib/vericite.rb
```diff
@@ -296,26 +301,26 @@ def sendRequest(command, args)
...
             end
           else
             # since we didn't have to consult VeriCite, set response status to 200
             response[:return_code] = 200
           end
 
           # the user score map shouldn't be empty now (either grabbed from the cache or VeriCite)
-          if user_score_map != nil
-            user_score_map.each do |key, score|
+          if users_score_map["#{user_id}"] != nil
```
ktgeek: this should be: unless users_score_map["#{user_id}"].nil?

lib/vericite.rb
```diff
@@ -281,13 +281,18 @@ def sendRequest(command, args)
...
           user_id = user.id
-          user_score_cache_key = "vericite_scores/#{consumer}/#{context_id}/#{user_id}"
-          # first check if the cache already has the score:
-          user_score_map = Rails.cache.read(user_score_cache_key)
-          if user_score_map == nil
+          user_score_cache_key_prefix =  "vericite_scores/#{consumer}/#{context_id}/#{assignment_id}/"
+          users_score_map = {}
+          # first check if the cache already has the user's score and if we haven't looked up this assignment lately:
+          users_score_map["#{user_id}"] = Rails.cache.read("#{user_score_cache_key_prefix}#{user_id}")
+          if users_score_map["#{user_id}"] == nil && Rails.cache.read(user_score_cache_key_prefix) == nil
```
ktgeek: variable.nil? is the preferred case in conditions in that if statement.

lib/vericite.rb
```diff
@@ -281,13 +281,18 @@ def sendRequest(command, args)
...
+            #we already looked up this user in Redis, don't bother again (by setting {})
+            users_score_map["#{user_id}"] ||= {}
             # we need to look up the user scores in VeriCite for this course
             # @return [Array<ReportScoreReponse>]
-            data, status_code, headers = vericite_client.reports_scores_context_id_get(context_id, consumer, consumer_secret, {:'user_id' => user_id})
+            data, status_code, headers = vericite_client.reports_scores_context_id_get(context_id, consumer, consumer_secret, {:'assignment_id' => assignment_id})
```
ktgeek: Either just `:assignment_id => assignment_id` or `assignment_id: assignment_id`  The later is the preferred form these days.

ktgeek: Also, it looks like the headers variable is assigned but never used.  (In a few places in here.)  You can use the variable name of `_` or `_headers` as a clue to other devs.  (Our linter caught this one.)

ktgeek:
This code doesn't help with the speed problem we're discussion out of band.  Just leaving a comment here to note that its not hitting a goal being discussed outside of GitHub.

ktgeek:
Passed internal tests and my manual QA, looks like we're good to merge

simonista:
Merged in 950a756f6fe7714ee599ef03259611a4c944451a

PR #1075, opened 2017-07-12
Paypal and Paytm 

GHanumanthaRayudu:
* Integrated paypal payment gateway
* integrated paytm payment gateway


jenseng:
Hi @GHanumanthaRayudu,

This feature doesn't make sense in the core canvas-lms repo. If this is something you need for your use case, we recommend that you implement it as a custom plugin that extends Canvas. This way it will be easier to maintain it in the future, as Canvas continues to change. For an example of how to enhance canvas, look at the [analytics plugin](https://github.com/instructure/analytics). The plugin mechanism allows you to enhance existing models/views/controllers, or add your own as the case may be.

Also, it looks like you checked in your payment credentials (see the config/environments files). You should get new credentials issued ASAP, as these ones have been exposed to the world, and there's no way to know who might already have them.

PR #1079, opened 2017-07-21
BigBlueButton feature: Delete recordings individually

jfederico:
Summary:
Implementation of delete recordings that can be performed on BigBlueButton Conferences when they hold recordings.

  Description:
The modification to the UI is based on the Mockup shared by Matt Goodwin. The icon used for the external link is the same used before so we don't affect themes.

As this feature is particular for BBB, a new attribute (:vendor => 'bigbluebuton') was added to the recording the code so in handlebars the UI can be differentiated.

There is some additional code in coffee script for ensuring that the action (delete) was performed. This is because when BBB is behind a load balancer (depending of the implementation) there may be a delay in its execution. This confirmation will show the user that the action is in process (disabling the recording link) and performing delete when is confirmed. If delete fails, a message is shown and the recording is restored in the UI (both, the button and the link)

 Test Plan:
   - As an instructor or administrator create a conference and join the session. Click on the recording button and after a few seconds logout from BBB.
   - The recording has to be processed, so wait for some minutes and refresh the page. More recordings can be created in the meanwhile following step 1.
   - When the recording is ready, the conference (long lasting running or ended) title can be expanded. The recording (or recordings) should be shown, one per row. At the left the new button "Delete" should be shown.
   - Click on delete and the delete message will be sent to BBB. An spinning wheel may be shown for a sec (or more if the delete is not confirmed in the attempt) and the row is deleted.

When the spinning wheel is shown for more than 1 sec means that the delete was enqueued but needs to be confirmed, in such case there will be 5 attempts of confirmation. If the delete is not confirmed, the row will be restored (link and button) and a flash message will be shown.

### DIFF
--- a/app/coffeescripts/handlebars_helpers.coffee
+++ b/app/coffeescripts/handlebars_helpers.coffee
@@ -150,6 +150,30 @@ define [
       real_min_str = (if real_minutes < 10 then "0" + real_minutes else real_minutes)
       "#{hours}:#{real_min_str}"
 
+    ###*
+     * Convert the total amount of minutes into a readable duration.
+     * @param {number}  Duration in minutes elapsed
+     * @return {string} String containing a formatted duration including hours and minutes
+     * Example:
+     *     ...
+     *     duration = 97
+     *     durationToString(duration)
+     *     ...
+     *     Returns
+     *       "Duration: 1 hour and 37 minutes"
+    ###
+    durationToString : (duration) ->
+      # stores the hours in the duration
+      hours = Math.floor(duration / 60)
+      # stores the remaining minutes after substracting the hours
+      minutes = duration % 60
+      if hours > 0
+        return I18n.t("Duration: %{hours} hours and %{minutes} minutes", {hours: hours, minutes: minutes})
+      else if minutes > 1
+        return I18n.t("Duration: %{minutes} minutes", {minutes: minutes})
+      else
+        return I18n.t("Duration: 1 minute")
+
     # helper for easily creating icon font markup
     addIcon : (icontype) ->
       new Handlebars.SafeString "<i class='icon-#{htmlEscape icontype}'></i>"
--- a/app/coffeescripts/views/conferences/ConferenceView.coffee
+++ b/app/coffeescripts/views/conferences/ConferenceView.coffee
@@ -20,9 +20,10 @@ define [
   'jquery'
   'Backbone'
   'jst/conferences/newConference'
+  'str/htmlEscape'
   'jquery.google-analytics'
   '../../jquery.rails_flash_notifications'
-], (I18n, $, {View}, template) ->
+], (I18n, $, {View}, template, htmlEscape) ->
 
   class ConferenceView extends View
 
@@ -38,6 +39,7 @@ define [
       'click .close_conference_link': 'close'
       'click .start-button': 'start'
       'click .external_url': 'external'
+      'click .delete_recording_link': 'deleteRecording'
 
     initialize: ->
       super
@@ -126,3 +128,46 @@ define [
         else
           window.open(data[0].url)
       )
+
+    deleteRecording: (e) ->
+      e.preventDefault()
+      $deleteButton = $(e.currentTarget).parent('div.ig-button')
+      if !confirm I18n.t("Are you sure you want to delete this recording?")
+        $deleteButton.focus()
+        return
+      $.ajaxJSON($deleteButton.data('url') + "/recording", "DELETE", {
+          recording_id: $deleteButton.data("id"),
+        }
+      ).done( (data, status) =>
+        if data.deleted
+          return @removeRecordingRow($deleteButton)
+        $.flashError(I18n.t("Sorry, the action performed on this recording failed. Try again later"))
+      ).fail( (xhr, status) =>
+        $.flashError(I18n.t("Sorry, the action performed on this recording failed. Try again later"))
+      )
+
+    removeRecordingRow: ($button) =>
+      $row = $('.ig-row[data-id="' + $button.data("id") + '"]')
+      $conferenceId = $($row.parents('div.ig-sublist')).data('id')
+      $row.parents('li.recording').remove()
+      @updateConferenceDetails($conferenceId)
+      $.screenReaderFlashMessage(I18n.t('Recording was deleted'))
+
+    updateConferenceDetails: (id) =>
+      $info = $('div.ig-row#conf_' + id).find('div.ig-info')
+      $detailRecordings = $info.find('div.ig-details__item-recordings')
+      $recordings = $('.ig-sublist#conference-' + id)
+      recordings = $recordings.find('li.recording').length
+      if recordings > 1
+        $detailRecordings.text(I18n.t("%{count} Recordings", {count: recordings}))
+        return
+      if recordings == 1
+        $detailRecordings.text(I18n.t("%{count} Recording", {count: 1}))
+        return
+      $detailRecordings.remove()
+      $recordings.remove()
+      # Shift the link to text
+      $link = $info.children('a.ig-title')
+      $text = $('<span />').addClass('ig-title').html($link.text())
+      $info.prepend($text)
+      $link.remove()
--- a/app/controllers/conferences_controller.rb
+++ b/app/controllers/conferences_controller.rb
@@ -347,6 +347,27 @@ class ConferencesController < ApplicationController
     end
   end
 
+  def recording
+    if authorized_action(@conference, @current_user, :read)
+      recording = @conference.recordings.find{ |r| r[:recording_id] == params[:recording_id] }
+      @response = recording || {}
+      respond_to do |format|
+        format.html { redirect_to named_context_url(@context, :context_conferences_url) }
+        format.json { render :json => @response }
+      end
+    end
+  end
+
+  def delete_recording
+    if authorized_action(@conference, @current_user, :delete)
+      @response = @conference.delete_recording(params[:recording_id])
+      respond_to do |format|
+        format.html { redirect_to named_context_url(@context, :context_conferences_url) }
+        format.json { render :json => @response, :status => :ok }
+      end
+    end
+  end
+
   protected
 
   def require_config
--- a/app/views/jst/conferences/concludedConference.handlebars
+++ b/app/views/jst/conferences/concludedConference.handlebars
@@ -7,15 +7,16 @@
             aria-controls="conference-{{id}}"
             aria-expanded="false"
             role="button">
-          <i class="icon-mini-arrow-right"></i>
+          <i class="icon-mini-arrow-right auto_rotate"></i>
           {{title}}
         </a>
       {{else}}
         <span class="ig-title">{{title}}</span>
       {{/if}}
+
       <div class="ig-details">
         <div class="ig-details__item">
-          {{dateString ended_at}}
+          {{datetimeFormatted ended_at}}
         </div>
         {{#if description}}
           <div class="ig-details__item ig-details__item--wrap-text">
@@ -23,12 +24,12 @@
           </div>
         {{/if}}
         {{#if recording}}
-          <div class="ig-details__item">
-            {{#if multipleRecordings}}
-              {{#t "recordings"}}{{recordingCount}} Recordings{{/t}}
-            {{else}}
-              {{#t "recording"}}1 Recording{{/t}}
-            {{/if}}
+          <div class="ig-details__item-recordings ig-details__item-recordings--wrap-text">
+          {{#if multipleRecordings}}
+            {{#t "recordings"}}{{recordingCount}} Recordings{{/t}}
+          {{else}}
+            {{#t "recording"}}1 Recording{{/t}}
+          {{/if}}
           </div>
         {{/if}}
       </div>
@@ -56,27 +57,31 @@
   </div>
 </div>
 {{#if recording}}
-  <div class="ig-sublist" id="conference-{{id}}" style="display: none;">
+  <div class="ig-sublist" data-id="{{id}}" id="conference-{{id}}" style="display: none;">
     <ul>
       {{#each recordings}}
-        <li>
-          <div class="ig-row">
+        <li class="recording">
+          <div class="ig-row" data-id="{{recording_id}}" style="text-align: center;">
             <div class="ig-row__layout">
               <div class="ig-info">
                 <div class="ig-details">
-                  <a href="{{playback_url}}" target="_blank" class="ig-title">
+                  <a href="{{playback_url}}" target="_blank" class="ig-title"
+                    style="line-height: inherit;" data-id="{{recording_id}}" data-format="presentation">
                     {{title}}
                   </a>
                   {{dateString created_at}}
                   &nbsp;&#124;&nbsp;
-                  {{minutesToHM duration_minutes}}
+                  {{durationToString duration_minutes}}
                 </div>
               </div>
-              <div class="ig-button">
-                <a class="btn btn-small" target="_blank" href="{{playback_url}}">
-                  {{#t "view"}}View{{/t}}
-                </a>
-              </div>
+              {{#if ../permissions.delete}}
+                <div class="ig-button" data-id="{{recording_id}}" data-action="delete" data-url="{{../../url}}">
+                  <a href="#"
+                      class="btn btn-small icon-trash delete_recording_link"
+                      title="{{#t "delete"}}Delete{{/t}}">{{#t "delete"}}Delete{{/t}}
+                  </a>
+                </div>
+              {{/if}}
             </div>
           </div>
         </li>
--- a/app/views/jst/conferences/newConference.handlebars
+++ b/app/views/jst/conferences/newConference.handlebars
@@ -7,7 +7,7 @@
             aria-controls="conference-{{id}}"
             aria-expanded="false"
             role="button">
-          <i class="icon-mini-arrow-right"></i>
+          <i class="icon-mini-arrow-right auto_rotate"></i>
           {{title}}
         </a>
       {{else}}
@@ -22,14 +22,20 @@
         {{#if scheduled}}
           <div>{{datetimeFormatted scheduled_at}}</div>
         {{/if}}
-        {{#if recording}}
-          {{#if multipleRecordings}}
-            <div>{{#t "recordings"}}{{recordingCount}} Recordings{{/t}}</div>
-          {{else}}
-            <div>{{#t "recording"}}1 Recording{{/t}}</div>
-          {{/if}}
+        {{#if description}}
+          <div class="ig-details__item ig-details__item--wrap-text">
+            {{description}}
+          </div>
         {{/if}}
-        <div>{{description}}</div>
+        {{#if recording}}
+          <div class="ig-details__item-recordings ig-details__item-recordings--wrap-text">
+          {{#if multipleRecordings}}
+            {{#t "recordings"}}{{recordingCount}} Recordings{{/t}}
+          {{else}}
+            {{#t "recording"}}1 Recording{{/t}}
+          {{/if}}
+          </div>
+        {{/if}}
       </div>
     </div>
     <div class="ig-admin">
@@ -85,27 +91,31 @@
   </div>
 </div>
 {{#if recording}}
-  <div class="ig-sublist" id="conference-{{id}}" style="display: none;">
+  <div class="ig-sublist" data-id="{{id}}" id="conference-{{id}}" style="display: none;">
     <ul>
       {{#each recordings}}
-        <li>
-          <div class="ig-row">
+        <li class="recording">
+          <div class="ig-row" data-id="{{recording_id}}" style="text-align: center;">
             <div class="ig-row__layout">
               <div class="ig-info">
                 <div class="ig-details">
-                  <a href="{{playback_url}}" target="_blank" class="ig-title">
+                  <a href="{{playback_url}}" target="_blank" class="ig-title"
+                    style="line-height: inherit;" data-id="{{recording_id}}" data-format="presentation">
                     {{title}}
                   </a>
                   {{dateString created_at}}
                   &nbsp;&#124;&nbsp;
-                  {{minutesToHM duration_minutes}}
+                  {{durationToString duration_minutes}}
                 </div>
               </div>
-              <div class="ig-button">
-                <a class="btn btn-small" target="_blank" href="{{playback_url}}">
-                  {{#t "view"}}View{{/t}}
-                </a>
-              </div>
+              {{#if ../permissions.delete}}
+                <div class="ig-button" data-id="{{recording_id}}" data-action="delete" data-url="{{../../url}}">
+                  <a href="#"
+                      class="btn btn-small icon-trash delete_recording_link"
+                      title="{{#t "delete"}}Delete{{/t}}">{{#t "delete"}}Delete{{/t}}
+                  </a>
+                </div>
+              {{/if}}
             </div>
           </div>
         </li>
--- a/spec/fixtures/files/conferences/big_blue_button_get_recordings_one.xml
+++ b/spec/fixtures/files/conferences/big_blue_button_get_recordings_one.xml
@@ -1,0 +1,39 @@
+<response>
+  <returncode>SUCCESS</returncode>
+  <recordings>
+    <recording>
+      <recordID>0225ccf234655ae60658ccac1e272d48781b491c-1511812307014</recordID>
+      <meetingID>instructure_web_conference_3Fn2k10wu0jK7diwJHs2FkDU0oXyX1ErUZCavikc</meetingID>
+      <internalMeetingID>0225ccf234655ae60658ccac1e272d48781b491c-1511812307014</internalMeetingID>
+      <name>John Doe's conference</name>
+      <isBreakout>false</isBreakout>
+      <published>true</published>
+      <state>published</state>
+      <startTime>1511812307014</startTime>
+      <endTime>1511812327102</endTime>
+      <participants>1</participants>
+      <metadata>
+        <meetingName>John Doe's conference</meetingName>
+        <meetingId>instructure_web_conference_3Fn2k10wu0jK7diwJHs2FkDU0oXyX1ErUZCavikc</meetingId>
+        <canvas-recording-ready-user>John Doe &lt;jdoe@mockmail.com&gt;</canvas-recording-ready-user>
+        <canvas-recording-ready-url>http://canvas.blah.com/mock/recording_ready</canvas-recording-ready-url>
+        <isBreakout>false</isBreakout>
+      </metadata>
+      <playback>
+        <format>
+          <type>presentation</type>
+          <url>http://bbb.blah.com/mock/playback/0225ccf234655ae60658ccac1e272d48781b491c-1511812307014</url>
+          <processingTime>9922</processingTime>
+          <length>0</length>
+          <preview>
+            <images>
+              <image alt="Welcome to" height="136" width="176">http://bbb.blah.com/mock/playback/0225ccf234655ae60658ccac1e272d48781b491c-1511812307014/d2d9a672040fbde2a47a10bf6c37b6a4b5ae187f-1511812307214/thumbnails/thumb-1.png</image>
+              <image alt="(this slide left blank for use as a whiteboard)" height="136" width="176">http://bbb.blah.com/mock/playback/0225ccf234655ae60658ccac1e272d48781b491c-1511812307014/d2d9a672040fbde2a47a10bf6c37b6a4b5ae187f-1511812307214/thumbnails/thumb-2.png</image>
+              <image alt="(this slide left blank for use as a whiteboard)" height="136" width="176">http://bbb.blah.com/mock/playback/0225ccf234655ae60658ccac1e272d48781b491c-1511812307014/d2d9a672040fbde2a47a10bf6c37b6a4b5ae187f-1511812307214/thumbnails/thumb-3.png</image>
+            </images>
+          </preview>
+        </format>
+      </playback>
+    </recording>
+  </recordings>
+</response>
--- a/spec/selenium/conferences/big_blue_button_conference_spec.rb
+++ b/spec/selenium/conferences/big_blue_button_conference_spec.rb
@@ -1,0 +1,105 @@
+#
+# Copyright (C) 2018 - present Instructure, Inc.
+#
+# This file is part of Canvas.
+#
+# Canvas is free software: you can redistribute it and/or modify it under
+# the terms of the GNU Affero General Public License as published by the Free
+# Software Foundation, version 3 of the License.
+#
+# Canvas is distributed in the hope that it will be useful, but WITHOUT ANY
+# WARRANTY; without even the implied warranty of MERCHANTABILITY or FITNESS FOR
+# A PARTICULAR PURPOSE. See the GNU Affero General Public License for more
+# details.
+#
+# You should have received a copy of the GNU Affero General Public License along
+# with this program. If not, see <http://www.gnu.org/licenses/>.
+
+require_relative '../common'
+require_relative '../helpers/conferences_common'
+require_relative '../helpers/public_courses_context'
+
+describe 'BigBlueButton conferences' do
+  include_context 'in-process server selenium tests'
+  include ConferencesCommon
+  include WebMock::API
+
+  bbb_endpoint = 'bbb.instructure.com'
+  bbb_secret = '8cd8ef52e8e101574e400365b55e11a6'
+  bbb_fixtures = {
+    :get_recordings => {
+      'meetingID' => 'instructure_web_conference_3Fn2k10wu0jK7diwJHs2FkDU0oXyX1ErUZCavikc',
+      'checksum' => 'ffa33deaa16e039bcbd84df90ed39196e518c69e'
+    },
+    :delete_recordings => {
+      'recordID' => '0225ccf234655ae60658ccac1e272d48781b491c-1511812307014',
+      'checksum' => 'f9a2b4c9b3e0dd611c1b541eee8bb32309a327ae'
+    }
+  }
+
+  before(:once) do
+    initialize_big_blue_button_conference_plugin bbb_endpoint, bbb_secret
+    course_with_teacher(name: 'Teacher Bob', active_all: true)
+    course_with_ta(name: 'TA Alice', course: @course, active_all: true)
+    4.times do |i|
+      course_with_student(name: "Student_#{i + 1}", course: @course, active_all: true)
+    end
+  end
+
+  before(:each) do
+    user_session(@teacher)
+    get conferences_index_page
+  end
+
+  after { close_extra_windows }
+  context 'when a conference is open' do
+    before(:once) do
+      stub_request(:get, /create/)
+    end
+
+    context 'and the conference has no recordings' do
+      before(:once) do
+        stub_request(:get, /getRecordings/).
+          with(query: bbb_fixtures[:get_recordings]).
+          to_return(:body => big_blue_button_mock_response('get_recordings', 'none'))
+        @conference = create_big_blue_button_conference(bbb_fixtures[:get_recordings]['meetingID'])
+      end
+
+      it 'should not include list with recordings', priority: '2' do
+        verify_conference_does_not_include_recordings
+      end
+    end
+
+    context 'and the conference has recordings' do
+      before(:once) do
+        stub_request(:get, /getRecordings/).
+          with(query: bbb_fixtures[:get_recordings]).
+          to_return(:body => big_blue_button_mock_response('get_recordings', 'two'))
+        @conference = create_big_blue_button_conference(bbb_fixtures[:get_recordings]['meetingID'])
+      end
+
+      it 'should include list with recordings', priority: '2' do
+        verify_conference_includes_recordings
+      end
+    end
+
+    context 'and the conference has one recording and it is deleted' do
+      before(:once) do
+        stub_request(:get, /deleteRecordings/).
+          with(query: bbb_fixtures[:delete_recordings]).
+          to_return(:body => big_blue_button_mock_response('delete_recordings'))
+        stub_request(:get, /getRecordings/).
+          with(query: bbb_fixtures[:get_recordings]).
+          to_return(:body => big_blue_button_mock_response('get_recordings', 'one'))
+        @conference = create_big_blue_button_conference(bbb_fixtures[:get_recordings]['meetingID'])
+      end
+
+      it 'should remove recording from the list', priority: '2' do
+        show_recordings_in_first_conference_in_list
+        delete_first_recording_in_first_conference_in_list
+        verify_conference_does_not_include_recordings
+      end
+    end
+
+  end
+end
--- a/app/models/big_blue_button_conference.rb
+++ b/app/models/big_blue_button_conference.rb
@@ -49,7 +49,7 @@ class BigBlueButtonConference < WebConference
     send_request(:create, {
       :meetingID => conference_key,
       :name => title,
-      :voiceBridge => "%020d" % self.global_id,
+      :voiceBridge => format("%020d", self.global_id),
       :attendeePW => settings[:user_key],
       :moderatorPW => settings[:admin_key],
       :logoutURL => (settings[:default_return_url] || "http://www.instructure.com"),
@@ -77,11 +77,11 @@ class BigBlueButtonConference < WebConference
     end
   end
 
-  def admin_join_url(user, return_to = "http://www.instructure.com")
+  def admin_join_url(user, _return_to = "http://www.instructure.com")
     join_url(user, :admin)
   end
 
-  def participant_join_url(user, return_to = "http://www.instructure.com")
+  def participant_join_url(user, _return_to = "http://www.instructure.com")
     join_url(user)
   end
 
@@ -90,10 +90,18 @@ class BigBlueButtonConference < WebConference
       recording_format = recording.fetch(:playback, {}).fetch(:format, {})
       {
         recording_id:     recording[:recordID],
+        title:            recording[:name],
         duration_minutes: recording_format[:length].to_i,
         playback_url:     recording_format[:url],
+        ended_at:         recording[:endTime].to_i,
       }
     end
+  end
+
+  def delete_recording(recording_id)
+    return { deleted: false } if recording_id.nil?
+    response = send_request(:deleteRecordings, recordID: recording_id)
+    { deleted: response.present? && response[:deleted].casecmp('true') == 0 }
   end
 
   def delete_all_recordings
@@ -112,7 +120,7 @@ class BigBlueButtonConference < WebConference
   def retouch?
     # If we've queried the room status recently, use that result to determine if
     # we need to recreate it.
-    if !@conference_active.nil?
+    unless @conference_active.nil?
       return !@conference_active
     end
 
@@ -120,7 +128,7 @@ class BigBlueButtonConference < WebConference
     # There's no harm in "creating" a room that already exists; the api will
     # just return the room info. So we'll just go ahead and recreate it
     # to make sure we don't accidentally redirect people to an inactive room.
-    return true
+    true
   end
 
   def join_url(user, type = :user)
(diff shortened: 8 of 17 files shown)


app/coffeescripts/views/conferences/ConferenceView.coffee
```diff
@@ -126,3 +128,63 @@ define [
...
+      parent = $(e.currentTarget).parent()
+      params = {recording_id: parent.data("id")}
+      this.toggleActionButton(parent, {state: "processing", action: "delete"})
+      this.toggleRecordingLink(parent, {state: "processing"})
+      $.ajaxJSON(parent.data('url') + "/delete_recording", "POST", params,
+        (data) =>
+          if data.deleted
+            this.removeRecordingRow(parent)
+            return
+          this.ensure_delete_performed({attempt: 1, parent: parent, desired_state: "true"})
```
simonista: why even pass `desired_state` if it's always true?

app/coffeescripts/views/conferences/ConferenceView.coffee
```diff
@@ -126,3 +128,63 @@ define [
...
+      )
+
+    ensure_delete_performed: (payload) ->
+      $.ajaxJSON(payload.parent.data('url') + "/recording", "POST", {
+          recording_id: payload.parent.data("id"),
+        }, (data) =>
+          current_state = if $.isEmptyObject(data) then "true" else "false"
+          if current_state == payload.desired_state
```
simonista: this should just collapse into `if $.isEmptyObject(data)`

app/controllers/conferences_controller.rb
```diff
@@ -347,6 +347,30 @@ def destroy
     end
   end
 
+  def recording
+    if authorized_action(@conference, @current_user, :delete)
+      @conference.transaction do
```
simonista: this doesn't need to be wrapped in a transaction

app/controllers/conferences_controller.rb
```diff
@@ -347,6 +347,30 @@ def destroy
...
+      respond_to do |format|
+        format.html { redirect_to named_context_url(@context, :context_conferences_url) }
+        format.json { render :json => @response }
+      end
+    end
+  end
+
+  def delete_recording
+    if authorized_action(@conference, @current_user, :delete)
+      @conference.transaction do
```
simonista: this doesn't need to be wrapped in a transaction

app/views/jst/conferences/concludedConference.handlebars
```diff
@@ -25,9 +25,9 @@
         {{#if recording}}
           <div class="ig-details__item">
             {{#if multipleRecordings}}
-              {{#t "recordings"}}{{recordingCount}} Recordings{{/t}}
+              <div>{{recordingCount}}&nbsp;{{#t "recordings"}}Recordings{{/t}}</div>
```
simonista: this is incorrect, the count needs to be inside the translation call

app/views/jst/conferences/concludedConference.handlebars
```diff
@@ -25,9 +25,9 @@
...
             {{else}}
-              {{#t "recording"}}1 Recording{{/t}}
+              <div>1&nbsp;{{#t "recording"}}Recording{{/t}}</div>
```
simonista: this is incorrect, the count needs to be inside the translation call

app/views/jst/conferences/concludedConference.handlebars
```diff
@@ -60,25 +60,55 @@
     <ul>
       {{#each recordings}}
         <li>
-          <div class="ig-row">
-            <div class="ig-row__layout">
-              <div class="ig-info">
-                <div class="ig-details">
-                  <a href="{{playback_url}}" target="_blank" class="ig-title">
-                    {{title}}
```
simonista: we don't want to show the title anyone where a BBB conference?

app/coffeescripts/views/conferences/ConferenceView.coffee
jfederico: Right, this comes from a more generic function we used before. The current function "ensure_delete_performed" used to be "ensure_action_performed". The validation was based on the parameter because it was used for different types of actions.

"published", "protected", etc

But now that we are only implementing delete, well it can be removed

app/coffeescripts/handlebars_helpers.coffee
```diff
@@ -143,6 +143,17 @@ define [
       real_min_str = (if real_minutes < 10 then "0" + real_minutes else real_minutes)
       "#{hours}:#{real_min_str}"
 
+    # Convert the total amount of minutes into a readable duration format.
+    durationToString : (minutes) ->
+      hours = Math.floor(minutes / 60)
+      real_minutes = minutes % 60
+      str = I18n.t('edit_conference_form.duration', 'Duration')+": "
+      if minutes>=60
```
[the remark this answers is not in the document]
jfederico: Right. I am applying the same rule to all the ifs. I noticed the original code does not have round braces. I know it is a matter of choice, do you guys prefer to use them or not?

app/coffeescripts/views/conferences/ConferenceView.coffee
jfederico: Indeed, current_state is actually not required here. Same reason as befor. The generic method in order to avoid duplication :)

app/coffeescripts/handlebars_helpers.coffee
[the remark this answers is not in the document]
simonista: I think in Coffeescript we prefer not to have parens around if statements.

app/views/jst/conferences/concludedConference.handlebars
jfederico: Well, the title has actually never been shown. I didn't go further because the recording title is always going to be the conference name anyways and the icon is not very explicit on what it is
image

app/views/jst/conferences/concludedConference.handlebars
jfederico: I'll resore the previous code

app/views/jst/conferences/concludedConference.handlebars
jfederico: I'll resore the previous code

app/coffeescripts/handlebars_helpers.coffee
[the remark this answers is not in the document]
jfederico: Got it.

app/coffeescripts/handlebars_helpers.coffee
```diff
@@ -143,6 +143,17 @@ define [
       real_min_str = (if real_minutes < 10 then "0" + real_minutes else real_minutes)
       "#{hours}:#{real_min_str}"
 
+    # Convert the total amount of minutes into a readable duration format.
+    durationToString : (minutes) ->
```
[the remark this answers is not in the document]
jfederico: Correct.

app/views/jst/conferences/concludedConference.handlebars
jfederico: There was a problem in the model. I added title based on recording.name

simonista: I see, yeah, it looks like the title was coming through blank.

It'd be great to not have to have a big if/else on vendor with lots of duplicated code, but instead just add the code for the delete button if it's available for that recording.

jfederico: Agreed!. That was my main concern one year ago when we started with the first PR. I don't know if you remember, but there was this massive PR with about 10 new features :laughing: 

I honestly don't remember to whom (from Canvas) we were speaking with at that time, but we came to the conclusion that we were going to need to do a separation of code as BBB is offering functionality that no other is. The if was the only thing we had handy as it was the less complicated.

But yeah, we can do this kind of simplified implementation for the delete PR. Although sooner o later we are going to need to separate the code. I still have a long list of features (and the code somehow ready), but the agreement with Matt Goodwin is that we will do small PRs one at the time :smile:

jfederico:
@simonista I applied the last changes requested on the handlebars files so there is no longer a vendor validation.

I also forced a push to tmy repo so the PR is cleaner, with only one commit to review/merge.

If you see anything else please let me know.

Thanks,

jfederico:
@simonista Just fixed a minor issue that came up after removing the vendor validation. The code should be good now for the last review (I hope :))

simonista:
Hey @jfederico, I'll get this pulled into our internal code review system, run through all our tests, and ping the team in charge of conference integrations.  They'll probably have more feedback for you.

app/views/jst/conferences/concludedConference.handlebars
```diff
@@ -27,7 +27,7 @@
             {{#if multipleRecordings}}
               {{#t "recordings"}}{{recordingCount}} Recordings{{/t}}
             {{else}}
-              {{#t "recording"}}1 Recording{{/t}}
+              <div>1&nbsp;{{#t "recording"}}Recording{{/t}}</div>
```
simonista: here's another case where 1 needs to go inside the translate call

app/controllers/conferences_controller.rb
```diff
@@ -347,6 +347,26 @@ def destroy
     end
   end
 
+  def recording
+    if authorized_action(@conference, @current_user, :delete)
```
simonista: should this be :read instead of :delete?

app/views/jst/conferences/newConference.handlebars
```diff
@@ -22,14 +22,18 @@
...
+            {{description}}
+          </div>
+        {{/if}}
         {{#if recording}}
           {{#if multipleRecordings}}
-            <div>{{#t "recordings"}}{{recordingCount}} Recordings{{/t}}</div>
+            {{#t "recordings"}}{{recordingCount}} Recordings{{/t}}
           {{else}}
-            <div>{{#t "recording"}}1 Recording{{/t}}</div>
+            <div>1&nbsp;{{#t "recording"}}Recording{{/t}}</div>
```
simonista: here's another case where 1 needs to go inside the translate call

app/views/jst/conferences/concludedConference.handlebars
```diff
@@ -60,23 +60,29 @@
...
+                  <a href="{{playback_url}}" target="_blank" class="ig-title"
+                    style="line-height: inherit;" data-id="{{recording_id}}" data-format="presentation">
                     {{title}}
                   </a>
-                  {{dateString created_at}}
-                  &nbsp;&#124;&nbsp;
-                  {{minutesToHM duration_minutes}}
+                  {{dateString ended_at}}
+                  &#124;&nbsp;
+                  {{durationToString duration_minutes}}
```
simonista: can you say more about these changes?  why change from `created_at` -> `ended_at` and why change the format of the duration time string?

simonista:
Hey @jfederico, just realized that this has no automated tests. Is that something you could add? I can't remember if we've talked about that already. Let me know if you'd like some additional guidance on what tests would be appropriate.

app/views/jst/conferences/newConference.handlebars
```diff
@@ -22,14 +22,18 @@
...
           <div>{{datetimeFormatted scheduled_at}}</div>
         {{/if}}
+        {{#if description}}
+          <div class="ig-details__item ig-details__item--wrap-text">
+            {{description}}
+          </div>
+        {{/if}}
         {{#if recording}}
           {{#if multipleRecordings}}
-            <div>{{#t "recordings"}}{{recordingCount}} Recordings{{/t}}</div>
```
simonista: why delete the `<div>` tags around this?

bootstraponline:
> This branch has conflicts that must be resolved
> Use the web editor or the command line to resolve conflicts.

bootstraponline:
I think we'd want a WebDriver UI test to verify the delete function works as expected. The network requests can be mocked for speed/stability. I don't think we have resources on the Instructure side to be running manual test cases for 3rd party features.

spencerolson:
Hi @jfederico  , thanks for your contribution! Your code is being reviewed by our developers. Once the review is complete, I'll A) post code review comments here if changes are requested, or B) merge the code if it is deemed ready-to-merge.

Thanks,
Spencer

app/coffeescripts/views/conferences/ConferenceView.coffee
```diff
@@ -126,3 +128,81 @@ define [
         else
           window.open(data[0].url)
       )
+
+    deleteRecording: (e) ->
+      return if !confirm(I18n.t('recordings.confirm.delete', "Are you sure you want to delete this recording?"))
```
spencerolson: We don't use keys anymore with I18n so we can remove `'recordings.confirm.delete',` here.

app/coffeescripts/views/conferences/ConferenceView.coffee
```diff
@@ -126,3 +128,81 @@ define [
...
+          recording_id: $button.data("id"),
+        }, (data) =>
+          if $.isEmptyObject(data)
+            callback($button)
+            return
+          if attempt < 5
+            attempt += 1
+            setTimeout((=> @ensureActionPerformed($button, attempt, callback); return;), attempt * 1000)
+            return
+          $.flashError(I18n.t('conferences.recordings.action_error', "Sorry, the action performed on this recording failed. Try again later"))
```
spencerolson: We don't use keys anymore with I18n so we can remove `'conferences.recordings.action_error',` here.

app/coffeescripts/views/conferences/ConferenceView.coffee
```diff
@@ -126,3 +128,81 @@ define [
...
+      @updateConferenceDetails(containerId.substring(11, containerId.length))
+
+    updateConferenceDetails: (id) =>
+      $info = $('div.ig-row#conf_' + id).children().children('div.ig-info')
+      $detailRecordings = $info.children('div.ig-details').children('div.ig-details__item-recordings')
+      $recordings = $('.ig-sublist#conference-' + id)
+      recordings = $recordings.children().children().length
+      # If it has more than one recording
+      if recordings > 1
+        $detailRecordings.text(I18n.t('recordings.recordings', "%{count} Recordings", {count: recordings}))
```
spencerolson: See comments above; same thing.

app/coffeescripts/views/conferences/ConferenceView.coffee
```diff
@@ -126,3 +128,81 @@ define [
...
+        return
+      # If it has only one recording
+      if recordings == 1
+        $detailRecordings.text(I18n.t('recordings.recording', "%{count} Recording", {count: 1}))
```
spencerolson: See comments above; same thing.

app/coffeescripts/views/conferences/ConferenceView.coffee
```diff
@@ -126,3 +128,81 @@ define [
...
+        return
+      # If it has no recordings
+      $detailRecordings.remove()
+      $recordings.remove()
+      @shiftLinkToText($info, 'ig-title')
+
+    shiftLinkToText: ($container, target) =>
+      $link = $container.children('a.' + target)
+      $container.prepend('<span class="' + target + '">' + $link.text() + '</span>')
```
spencerolson: our XSS linter flagged this as possibly XSS-able, and caused the build to fail. I think we're safe here because `target` is not coming from user input, and JQuery's .text() should be XSS-safe. Can you confirm that this is XSS-safe and add a comment above this line to flag it as XSS-safe? https://github.com/jenseng/xsslint

I believe the comment should look something like:

  `# xsslint safeString.identifier target $link`

app/coffeescripts/views/conferences/ConferenceView.coffee
```diff
@@ -20,9 +20,10 @@ define [
   'jquery'
```
spencerolson: Do we not have any javascript specs around these ConferenceView changes? Would be nice to get some unit tests here, and if those new unit tests cover what is being currently tested by some of the selenium specs, remove those selenium specs.

spec/selenium/conferences/big_blue_button_conference_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+        stub_request(:get, 'bbb.instructure.com/bigbluebutton/api/deleteRecordings').
+        with(query: bbb_fixtures[:delete_recordings]).
+        to_return(:body => big_blue_button_mock_response('delete_recordings'))
+        stub_request(:get, 'bbb.instructure.com/bigbluebutton/api/getRecordings').
+        with(query: bbb_fixtures[:get_recordings]).
+        to_return(:body => big_blue_button_mock_response('get_recordings', 'one'))
+        @conference = create_big_blue_button_conference(bbb_fixtures[:get_recordings]['meetingID'])
+      }
+
+      it 'should remove recording from the list', priority: '2' do
```
spencerolson: This test is failing.

```
 BigBlueButton conferences when a conference is open and the conference has one recording and it is deleted should remove recording from the list
  Failure/Error: f('.delete_recording_link', new_conference_list)
  
  Selenium::WebDriver::Error::NoSuchElementError:
    Unable to locate element: :css, ".delete_recording_link"
```

spec/selenium/conferences/big_blue_button_conference_spec.rb
```diff
@@ -0,0 +1,104 @@
...
+
+    context 'and the conference has recordings' do
+      before(:once) {
+        stub_request(:get, 'bbb.instructure.com/bigbluebutton/api/getRecordings').
+        with(query: bbb_fixtures[:get_recordings]).
+        to_return(:body => big_blue_button_mock_response('get_recordings', 'two'))
+        @conference = create_big_blue_button_conference(bbb_fixtures[:get_recordings]['meetingID'])
+      }
+
+      it 'should include list with recordings', priority: '2' do
```
spencerolson: This test is failing.

```
BigBlueButton conferences when a conference is open and the conference has recordings should include list with recordings
  Failure/Error: expect(first_conference_in_list(new_conference_list)).to include_text 'Recording'
  
    expected #<Selenium::WebDriver::Element:0x..fc605d7237afed1aa id="{fe2a1da1-038f-40ae-b5a5-bcfbfcf5f381}"> text to include Recording, actual text was: BigBlueButton Conference
    Start
    Settings
```

app/coffeescripts/views/conferences/ConferenceView.coffee
jfederico: Well, the link is already there and it is originally rendered by the template, so yes it is safe, but I am now letting jquery to create the node. That is safer! :)

app/coffeescripts/views/conferences/ConferenceView.coffee
jfederico: Nope, there were no tests at all on conferences. I agree that more should be added. Although in this case I focused all the effort only in the new feature for big_blue_button as the Conference spec is actually testing Wimba (only).

I would suggest to complete this one and we can certainly help adding some more tests in a next PR if you wanted to.

spencerolson:
Hi @jfederico , I ran your most recent changes through tests again. Those two failing specs that I mentioned before are still failing (`spec/selenium/conferences/big_blue_button_conference_spec.rb:81`
and `spec/selenium/conferences/big_blue_button_conference_spec.rb:97`).

In addition, the reviewers have asked that you please add unit tests to `spec/coffeescripts/views/conferences/ConferenceViewSpec.coffee` to test the changes/additions introduced in `app/coffeescripts/views/conferences/ConferenceView.coffee`.

Once we get all tests passing, and have unit tests for `app/coffeescripts/views/conferences/ConferenceView.coffee`, I'll make sure this gets in front of our reviewers again.

Thanks!
Spencer

spec/selenium/conferences/big_blue_button_conference_spec.rb
jfederico: Yeah, for this one I needed to expand the recordings before executing the click on delete.
```
Finished in 7.67 seconds (files took 18.31 seconds to load)
3 examples, 0 failures

Randomized with seed 61250
```
Although the error you are pointing at makes me think that the test it is not using the correct assets.

Unable to locate element: :css, ".delete_recording_link"

For what other reason the the element .delete_recording_link would not be found. Would it be a  css required?

spec/selenium/conferences/big_blue_button_conference_spec.rb
jfederico: This one, I don't have idea why is happening. The text shown here:
```
expected #<Selenium::WebDriver::Element:0x..fc605d7237afed1aa id="{fe2a1da1-038f-40ae-b5a5-bcfbfcf5f381}"> text to include Recording, actual text was: BigBlueButton Conference
    Start
    Settings
```
If we were executing the opposite test (verify_conference_does_not_include_recordings) it should be like this:
```
 expected #<Selenium::WebDriver::Element:0x..fc484bc3484ccba84 id="4"> text to NOT include Recording, actual text was: BigBlueButton Conference
       2 Recordings
       Start
       Settings
       Conferencia de Development 101
       Links to an external site.
        |  Duration: 1 minute
```
Which makes me think that it is either (as for the previus one) that the assets used are not the new ones, or that webmock is not stubbing the request.

But here I don't know what to do as the local tests are passing
```
RAILS_ENV=test rake spec:selenium
```

spec/selenium/conferences/big_blue_button_conference_spec.rb
spencerolson: Hi @jfederico , our continuous integration process will compile assets, so I'm not sure if that's what the issue would be. My best guess is that the link is on the page but hidden, or not on the page at all. If this helps, here is a screenshot of the UI when the failure occurs:
screen shot 2018-02-12 at 3 40 04 pm

jfederico:
Hi @spencerolson,
Thanks for checking it out. Although I hadn't worked on the tests themselves :P 
I just pushed another update and added some comments after yours.

>In addition, the reviewers have asked that you please add unit tests to spec/coffeescripts/views/conferences/ConferenceViewSpec.coffee to test the changes/additions introduced in app/coffeescripts/views/conferences/ConferenceView.coffee.

Well, this is what I was afraid of, as we keep adding and adding requirements :(

It is all right tough, I'll be more than happy to help you guys out. Just a quick question, who would be the person requiring these tests?, any chance I can talk to he/she in Slack?

Thanks again for your help.

spec/selenium/conferences/big_blue_button_conference_spec.rb
spencerolson: Hi @jfederico I'll retrigger another pass of these tests to make sure we get the same failures. I can also try running these tests locally to see if I get any different results.

For reference, here's a screenshot of the page when the test is failing:
screen shot 2018-02-12 at 3 52 55 pm

spec/selenium/conferences/big_blue_button_conference_spec.rb
jfederico: Alright. If we discard the assets, the next thing to look up would be the fixtures. Is it possible that they are not getting there?
If they where, the UI sould look like this:

image

And expanding the recordings like this:

image

spec/selenium/conferences/big_blue_button_conference_spec.rb
spencerolson: OK, just ran this spec file locally after compiling assets, and my results agree with the build. I am seeing these two specs fail:

```
Failed examples:

rspec ./spec/selenium/conferences/big_blue_button_conference_spec.rb:97 # BigBlueButton conferences when a conference is open and the conference has one recording and it is deleted should remove recording from the list
rspec ./spec/selenium/conferences/big_blue_button_conference_spec.rb:81 # BigBlueButton conferences when a conference is open and the conference has recordings should include list with recordings

Randomized with seed 15803
```
Can you try running the specs with the seed set to 15803 (to match my run above)?

`rspec spec/selenium/conferences/big_blue_button_conference_spec.rb --seed 15803`

spencerolson:
Hi @jfederico apologies, I did not see that you had pushed up a new commit. I was testing against the prior version. Running the newest commit through tests now.

spec/selenium/conferences/big_blue_button_conference_spec.rb
jfederico: It turned out no to be an issue with the fixture files themselves but with the way the request was done. The model makes requests to BBB as https and it is hardcoded. The specs were considering http as it was not specified.

It should be good now. Still missing the test to coffescript.

spencerolson: Alright, great news! I took your newest commit and pushed it through tests, and they all passed. Once we have the Coffeescript tests for ConferenceView, I'll get this back in front of reviewers.

Thanks!

app/views/jst/conferences/concludedConference.handlebars
```diff
@@ -59,24 +60,32 @@
...
-                </a>
-              </div>
+              {{#if ../permissions.delete}}
+                <div class="ig-loader" data-id="{{recording_id}}" data-action="delete" style="display: none;">
+                  <img src="/images/ajax-loader-black-on-white.gif" class="loader" />
+                </div>
+                <div class="ig-button" data-id="{{recording_id}}" data-action="delete" data-url="{{../../url}}">
+                  <a href="#"
+                      class="btn btn-small icon-trash delete_recording_link"
+                      aria-label="{{#t "delete"}}Delete{{/t}}"
```
spencerolson: We shouldn't need this label.  Let the link describe what it actually is.  The label will make it more confusing.

app/views/jst/conferences/newConference.handlebars
```diff
@@ -88,24 +94,32 @@
...
-                </a>
-              </div>
+              {{#if ../permissions.delete}}
+                <div class="ig-loader" data-id="{{recording_id}}" data-action="delete" style="display: none;">
+                  <img src="/images/ajax-loader-black-on-white.gif" class="loader" />
+                </div>
+                <div class="ig-button" data-id="{{recording_id}}" data-action="delete" data-url="{{../../url}}">
+                  <a href="#"
+                      class="btn btn-small icon-trash delete_recording_link"
+                      aria-label="{{#t "delete"}}Delete{{/t}}"
```
spencerolson: We shouldn't need this label.  Let the link describe what it actually is.  The label will make it more confusing.

spec/fixtures/files/conferences/big_blue_button_get_recordings_one.xml
```diff
@@ -0,0 +1,39 @@
...
+              <image alt="Welcome to" height="136" width="176">http://bbb.instructure.com/presentation/0225ccf234655ae60658ccac1e272d48781b491c-1511812307014/presentation/d2d9a672040fbde2a47a10bf6c37b6a4b5ae187f-1511812307214/thumbnails/thumb-1.png</image>
+              <image alt="(this slide left blank for use as a whiteboard)" height="136" width="176">http://bbb.instructure.com/presentation/0225ccf234655ae60658ccac1e272d48781b491c-1511812307014/presentation/d2d9a672040fbde2a47a10bf6c37b6a4b5ae187f-1511812307214/thumbnails/thumb-2.png</image>
+              <image alt="(this slide left blank for use as a whiteboard)" height="136" width="176">http://bbb.instructure.com/presentation/0225ccf234655ae60658ccac1e272d48781b491c-1511812307014/presentation/d2d9a672040fbde2a47a10bf6c37b6a4b5ae187f-1511812307214/thumbnails/thumb-3.png</image>
+            </images>
+          </preview>
+        </format>
+      </playback>
+    </recording>
+  </recordings>
+</response>
```
bootstraponline: Lets make sure all the urls in here are not real urls

jfederico: Done

jfederico:
Hey there. I just applied the latest changes required to make sure none of the URL is a real URL

app/coffeescripts/handlebars_helpers.coffee
```diff
@@ -150,6 +150,16 @@ define [
...
+      hours = Math.floor(minutes / 60)
+      real_minutes = minutes % 60
```
omarkhan: How come you are passing in minutes but then you have to convert to real minutes? Do you mean that the param is actually seconds? I would reword this variable because this doesn't make sense to me. I would also add some comments on what is getting passed in and document the math you are doing

app/coffeescripts/views/conferences/ConferenceView.coffee
```diff
@@ -126,3 +128,83 @@ define [
         else
           window.open(data[0].url)
       )
+
+    deleteRecording: (e) ->
+      e.preventDefault()
+      if !confirm I18n.t("Are you sure you want to delete this recording?")
+        $(e.currentTarget).parents('.inline-block').find('.al-trigger').focus()
+        return
```
omarkhan: So if I understand this you are seeing if they haven't confirmed to delete the recording you pop a modal with that text then from there you put focus on what? I would like a comment here to tell us what is that focus otherwise we will be sad for a11y.

app/coffeescripts/views/conferences/ConferenceView.coffee
```diff
@@ -126,3 +128,83 @@ define [
...
+      $deleteButton = $(e.currentTarget).parent()
```
omarkhan: This is very magical.  When you put focus on some parents we are lost as to what that is.  It would be nice to tell us in a comment since this is not self documenting really.

app/coffeescripts/views/conferences/ConferenceView.coffee
```diff
@@ -126,3 +128,83 @@ define [
...
+      @toggleActionButton($deleteButton, {state: "processing", action: "delete"})
+      @toggleRecordingLink($deleteButton, {state: "processing"})
+      $.ajaxJSON($deleteButton.data('url') + "/recording", "DELETE", {
+          recording_id: $deleteButton.data("id"),
+        }, (data) =>
+          if data.deleted == "true"
+            @removeRecordingRow($deleteButton)
+            return
+          @ensureActionPerformed($deleteButton, 1, @removeRecordingRow)
+      )
```
omarkhan: I mean this is okay I guess but I would rather have a .then and catch rather than just a data that returns that we check for true but I won't -1 on that.

app/coffeescripts/views/conferences/ConferenceView.coffee
```diff
@@ -126,3 +128,83 @@ define [
...
+
+    ensureActionPerformed: ($button, attempt, callback) =>
+      $.ajaxJSON($button.data('url') + "/recording", "GET", {
```
omarkhan: Should you not be able to get this from the ENV? why are you grabbing the url from the button? I assume you are grabbing all the recordings again from the course it would be better to grab it from the ENV rather than using jquery to grab the url from a button.

app/coffeescripts/views/conferences/ConferenceView.coffee
```diff
@@ -126,3 +128,83 @@ define [
...
+          recording_id: $button.data("id"),
+        }, (data) =>
+          if $.isEmptyObject(data)
+            callback($button)
+            return
+          if attempt <= 5
+            attempt++
+            setTimeout((=> @ensureActionPerformed($button, attempt, callback); return;), attempt * 1000)
+            return
```
omarkhan: So if I understand this you are going to loop 5 times ensuring that we actually deleted until its actually deleted? Why is this necessary? it should just be one you are going to kill the server if we hit this 5 times if something fails.

app/coffeescripts/views/conferences/ConferenceView.coffee
```diff
@@ -126,3 +128,83 @@ define [
...
+      id = containerId.substring(11, containerId.length)
+      @updateConferenceDetails(containerId.substring(11, containerId.length))
+      $.screenReaderFlashMessage(I18n.t('Recording was deleted'))
+
+    updateConferenceDetails: (id) =>
+      $info = $('div.ig-row#conf_' + id).children().children('div.ig-info')
+      $detailRecordings = $info.children('div.ig-details').children('div.ig-details__item-recordings')
+      $recordings = $('.ig-sublist#conference-' + id)
+      recordings = $recordings.children().children().length
+      # If it has more than one recording
```
omarkhan: The code below is self documenting enough we don't need this comment

app/coffeescripts/views/conferences/ConferenceView.coffee
```diff
@@ -126,3 +128,83 @@ define [
...
+      if recordings > 1
+        $detailRecordings.text(I18n.t("%{count} Recordings", {count: recordings}))
+        return
+      # If it has only one recording
```
omarkhan: Same

app/coffeescripts/views/conferences/ConferenceView.coffee
```diff
@@ -126,3 +128,83 @@ define [
...
+      if recordings == 1
+        $detailRecordings.text(I18n.t("%{count} Recording", {count: 1}))
+        return
+      # If it has no recordings
```
omarkhan: Same

app/coffeescripts/handlebars_helpers.coffee
jfederico: Yeah, agree. It is kind of confusing in the way it is right now. And although it is based on the function written a few lines above (https://github.com/instructure/canvas-lms/blob/stable/app/coffeescripts/handlebars_helpers.coffee#L147) rephrasing and adding comments may help to make it clearer.

app/coffeescripts/views/conferences/ConferenceView.coffee
jfederico: Yes, this was a copy-paste from the delete method above. But it is not focusing on what it should. It may actually not be necessary. I'll double check the coffeescript test. Update the identifier if needed or remove the line.

app/coffeescripts/views/conferences/ConferenceView.coffee
jfederico: Right. The button is enclosed in a div element, would it help to add a selector? 

```$deleteButton = $(e.currentTarget).parent('.ig-button')```

or even

```$deleteButton = $(e.currentTarget).parent('div.ig-button')```

 Or an explicit comment would be preferred?

Sorry pal, I am just trying to understand what would make things easier for Engineers to review. I know the person reviewing is not always the same, so I am trying to find middle ground while being consistent with the code and styling in a clearer way.

File: app/coffeescripts/views/conferences/ConferenceView.coffee (hunk 32 above)
Comment (jfederico) [reply app/coffeescripts/views/conferences/ConferenceView.coffee]: Done

File: app/coffeescripts/views/conferences/ConferenceView.coffee (hunk 33 above)
Comment (jfederico) [reply app/coffeescripts/views/conferences/ConferenceView.coffee]: Done

File: app/coffeescripts/views/conferences/ConferenceView.coffee (hunk 30 above)
Comment (jfederico) [reply app/coffeescripts/views/conferences/ConferenceView.coffee]: Ok, a little bit of context.

The recordings are actually not stored in anyway in Canvas. Their metadata is retrieved from the BBB server with a getRecordings request per conference in the list. Yes, it may be a killer, but it is the way it has been implemented for ages, and it is the best way as you don't want to get in storing in Canvas and then syncing the content. As the recordings live in BBB, all the URLs point really to external links.

Also, every request for managing the recording (right now deleteRecordings but later on may be publishRecordings, updateRecordings and protectRecordings) has to be done through Canvas wher the endpoint is the conference itself, but there must be a reference to the recordingID (which is internal to BBB). 

So, the way join, close and even delete is performed on the conference URL (/courses/1/conferences/20/close), see:
```<a class="btn btn-small close_conference_link" href="/courses/1/conferences/20/close">End</a>```

would be the same way required for managing recordings

```
<div class="ig-button" data-id="162c30feb94c7dc336bfe075427f92f974fffd13-1516998778542" data-action="delete" data-url="/courses/1/conferences/20">
  <a href="#" class="btn btn-small icon-trash delete_recording_link" title="Delete">Delete</a>
</div>
```
As the request is performed with ajax (explanation in the next comment), both the URL and the ID have to be stored somewhere.

The recordingID corresponds to the recording, but the URL corresponds to the conference. It may be better then to store the URL at the conference level, but it is irrelevant if the conference doesn't have recordings. It can also be stored at the button link level, but then we would be repeating the same data for all the other actions (hypothetically, as the may be implemented in the near future)

File: app/coffeescripts/views/conferences/ConferenceView.coffee (hunk 31 above)
Comment (jfederico) [reply app/coffeescripts/views/conferences/ConferenceView.coffee]: First lets go to the reasons behind. Why ajax and no a direct request to the controller as the delete conference action?

As the recordings live in an external server (lets say BBB), the action deleteRecordings can fail. 

But in fact, the real issue is that this was required to workaround a delay in BBB for executing some of the actions. 

Until 4 months ago (the original version of this code was written more than 1 year ago), there was a delay when performing some actions on recordings. Which means that when the action was sent to BBB it was put in a queue and the BBB API doesn't support callbacks. As the action was not made effective immediately, performing a direct request to the controller would have lead to a new query for fetching the recordings from BBB resulting in a race condition that may have as a response a recording in the state it was before the user clicked the button. From the user perspective, the application was failing.

Doing it with ajax and performing an extra (or a few) requests to confirm that the expected state was actually reached is highly recommended (from us) to anyone doing integrations with BBB.

Now, why 5 times. It is really an arbitrary number. It could be 1 it could be 3, it didn't really matter as long as there was a mechanism for confirmation in place. This implementation performs the first verification after 1 sec, the second after two and so on. Usually the first one would give the correct state, but sometimes 3. 5 was more like to play 'safe'.

Now, putting this in context. Management action on recordings can be done only by the conference owners and administrators. So the possibility of over killing a server with multiple requests is really low, as not even them are performing deleteRecording all the time. 

To be honest, we may remove the verification now as the action is expected to be executed immediately. But I kept it because the recording still lives in an external server and there is always a chance of the action to fail. And as BBB doesn't have callbacks, well the only way to verify if the action was performed is to perform the second request.

So, what should we do?

File: app/coffeescripts/views/conferences/ConferenceView.coffee (hunk 29 above)
Comment (jfederico) [reply app/coffeescripts/views/conferences/ConferenceView.coffee]: I see what you mean. The only point of failure for the client is if Canvas goes down. This is because the controller/model are taking care of giving a response true/false to this ajax request.

What we can do to make it more resilient is to add a fail method with the corresponding code for failing over. Would that be OK?

**Comment by jfederico** [comment]
Guys, one more question. I run eslint and rlint (standard scripts) and I am not seeing anything wrong there.
```
ubuntu@canvas-dev:~/canvas-lms.test$ script/rlint 
-- -- -- -- -- -- -- -- -- -- --
No relevant Rubocop errors found!
-- -- -- -- -- -- -- -- -- -- --
ubuntu@canvas-dev:~/canvas-lms.test$ script/eslint 
-- -- -- -- -- -- -- -- -- -- --
No relevant eslint errors found!
-- -- -- -- -- -- -- -- -- -- --
```
But, with rubocop I can see a number of issues that even though are not related to the new code, they will show up in CI reports.
```
app/models/big_blue_button_conference.rb:44:29: C: Performance/TimesMap: Use Array.new(8) with a block instead of .times.map.
      settings[:user_key] = 8.times.map{ chars[chars.size * rand] }.join
                            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
app/models/big_blue_button_conference.rb:45:30: C: Performance/TimesMap: Use Array.new(8) with a block instead of .times.map.
      settings[:admin_key] = 8.times.map{ chars[chars.size * rand] }.join until settings[:admin_key] && settings[:admin_key] != settings[:user_key]
                             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
app/models/big_blue_button_conference.rb:52:31: C: Style/FormatString: Favor format over String#%.
      :voiceBridge => "%020d" % self.global_id,
                              ^
app/models/big_blue_button_conference.rb:80:28: W: Lint/UnusedMethodArgument: Unused method argument - return_to. If it's necessary, use _ or _return_to as an argument name to indicate that it won't be used.
  def admin_join_url(user, return_to = "http://www.instructure.com")
                           ^^^^^^^^^
app/models/big_blue_button_conference.rb:84:34: W: Lint/UnusedMethodArgument: Unused method argument - return_to. If it's necessary, use _ or _return_to as an argument name to indicate that it won't be used.
  def participant_join_url(user, return_to = "http://www.instructure.com")
                                 ^^^^^^^^^
app/models/big_blue_button_conference.rb:107:7: C: Style/NegatedIf: Favor unless over if for negative conditions.
      deleted = response[:deleted] if !response.nil?
      ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
app/models/big_blue_button_conference.rb:128:5: C: Style/NegatedIf: Favor unless over if for negative conditions.
    if !@conference_active.nil? ...
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^
app/models/big_blue_button_conference.rb:136:5: C: Style/RedundantReturn: Redundant return detected.
    return true
    ^^^^^^
app/models/big_blue_button_conference.rb:180:7: C: Layout/CaseIndentation: Indent when as deep as case.
      when Net::HTTPSuccess
      ^^^^
app/models/big_blue_button_conference.rb:192:57: C: Style/SpecialGlobalVars: Prefer $ERROR_INFO from the stdlib 'English' module (don't forget to require it) over $!.
    logger.error "big blue button unhandled exception #{$!}"
                                                        ^^
app/models/big_blue_button_conference.rb:218:22: C: Style/EachWithObject: Use each_with_object instead of reduce.
      child_elements.reduce({}) do |hash, child|
                     ^^^^^^
```
What would be the expectation here? Should we proceed to clean up that code or leave it for a second PR just for that?

app/coffeescripts/views/conferences/ConferenceView.coffee
sdb1228: @jfederico This url though still points to canvas yes? or does ```$button.data('url')``` in this context point to the BBB server that is getting the recordings? In either case the comment here is suggesting that there is a better way to grab this URL rather than using jquery to grab the data url.  If in fact we grab this url client side and inject it into the button then thats fine grabbing this way but I figured there was more configuration with this button url.

app/coffeescripts/views/conferences/ConferenceView.coffee
sdb1228: @jfederico You are right its hard to strike a balance especially in coffeescript land.  I would much rather have a selector on this ```$deleteButton = $(e.currentTarget).parent('div.ig-button')``` would be my preferred way.

app/coffeescripts/views/conferences/ConferenceView.coffee
sdb1228: @jfederico That would be great!

app/coffeescripts/views/conferences/ConferenceView.coffee
jfederico: Done

app/coffeescripts/views/conferences/ConferenceView.coffee
jfederico: Alright, yes and no.
All the request are going to BBB through Canvas
Client->Canvas->BBB
The delete is requested only once.
deleteRecording (Client)->recording DELETE (Canvas controller)->deleteRecordings (BBB API)

The second request (therefore the one that is repeated) pull the current status.
ensureActionPerformed (Client)->recording GET (Canvas controller)->getRecordings (BBB API)

But is cool, lets change the approach. As the delete (and all the other action on recordings) are now executed immediately in BBB (no more queue for processing) I am removing the code that does the verification.

app/coffeescripts/views/conferences/ConferenceView.coffee
jfederico: That is correct. But even in that case as ENV has two arrays with conferences (concluded_conferences and current_conferences) I would need to pull the conference id from the upper level using jquery and then lookup for the URL in the arrays. It will result in a less efficient solution in my opinion.

It is doable, and it is actually done when removing the row from the DOM, but is more direct this way.

Should I proceed?

jfederico:
Recorded this video as for testing the feature. The quality on YouTube is really bad, but the source can be provided upon request.
https://youtu.be/XDjjcM3p8n0

Perthenia:
The good news is that the delete is working as designed. We do have two questions: are you supposed to be able to see the list of recordings in progress/New as well?
screen shot 2018-03-26 at 3 58 17 pm
There is another question in our slack channel - please check and reply

bootstraponline:
Merged. https://github.com/instructure/canvas-lms/commit/95998334cddedf70fd8e22593f9680734f823867

Issue #1101, opened 2017-08-25
Generate source maps in production? (Of course, yes!)


grahamb:
I'm creating this issue as a discussion; PR forthcoming. Paging @claydiffrient & @ryankshaw to the thread.

TL; DR: sourcemaps are awesome, and they should be available in production, but node/V8 may explode trying to generate them.

Sometimes you get an error reported from a user, and it's in front-end code, and you've got to try to reproduce it and debug. Dev environments are awesome, but they're not always an option; sometimes you've got to jump in and debug it in production, or a production-like environment.

Canvas _used_ to have a neat trick where you could pass in a URL parameter and get un-minified JS. With the move to Webpack, this is gone, and as such, great sadness happens when you try to debug Canvas JS in production. I mean, have you _tried_ to step through minified, obfuscated, TinyMCE code? 😭 😭 😭

Sourcemaps are the answer to this, right? Right! They're being [generated in development](https://github.com/instructure/canvas-lms/blob/0eef2ce4b03f380e7566f52cd427e6b06630cb50/frontend_build/baseWebpackConfig.js#L75), but not in production.

Sourcemaps aren't free, though; there's a cost incurred in generating them. Specifically, two costs: time, and compute resources.

### Time
In my testing on our build machine (RHEL6 VM, quad-core Xeon 3.33GHz, 14GB RAM), generating sourcemaps took a few minutes longer than building without them:

Without sourcemaps:
```
real	5m32.714s
user	13m58.905s
sys	0m20.079s
```

With sourcemaps:
```
real	7m49.296s
user	17m35.509s
sys	0m27.662s
```

So, about 2.5 to 3 minutes longer. Not too shabby. YMMV.

### Resource Usage

Here's where things go off the rails. My first pass at this was to change `frontend_build/baseWebpackConfig` and `webpack.production.config.js` to emit sourcemaps (specifically, the `source-map` option). Now, I'm currently [about a thousand miles]( https://www.google.com/maps/dir/40.7597411,-111.8961407/Simon+Fraser+University,+8888+University+Dr,+Burnaby,+BC+V5A+1S6,+Canada/@44.9310266,-121.9362568,6z/data=!3m1!4b1!4m9!4m8!1m1!4e1!1m5!1m1!1s0x548679c0857a2227:0x506c3f6e30b55b!2m2!1d-122.9198833!2d49.2780937) away from my servers, but I imagine that it was looking something like this:

[...]

PR #1102, opened 2017-08-26
Generate JavaScript sourcemaps in production

grahamb:
Resolves instructure/canvas-lms#1101

In production, generate separate sourcemap files so that bundles appear split into their original, un-obfuscated source files in browser devtools.

Note: `source-map` generation on large projects can cause node/v8 to exhaust its memory allocation (see webpack/webpack#1914, webpack/webpack#2157). This is the reason for the change to package.json. In the future (e.g. when Canvas uses Node 8, or the NODE_OPTIONS feature is back-ported to Node 6, this can go back to a traditional npm script that doesn't directly invoke `node`.

Test Plan:
- Remove the `public/dist/webpack-*` directories.
- Run `JS_BUILD_NO_UGLIFY=true npm run webpack-production`. No sourcemap files should be present in  `public/dist/webpack-production`.
- Run `npm run webpack-production`. Sourcemap files should be present in `public/dist/webpack-production`

### DIFF
--- a/package.json
+++ b/package.json
@@ -152,7 +152,7 @@
     "webpack": "gulp rev & webpack --progress --color --watch",
     "webpack-development": "gulp rev & webpack --progress --color",
     "webpack-stats": "webpack --profile --json",
-    "webpack-production": "NODE_ENV=production gulp rev & webpack --color --config webpack.production.config.js",
+    "webpack-production": "gulp rev & node --max_old_space_size=4096 node_modules/webpack/bin/webpack.js --progress --color --config webpack.production.config.js",
     "jspec": "./spec/jspec.sh",
     "jspec-watch": "./spec/jspec.sh --watch"
   }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -152,7 +152,7 @@
     "webpack": "gulp rev & webpack --progress --color --watch",
     "webpack-development": "gulp rev & webpack --progress --color",
     "webpack-stats": "webpack --profile --json",
-    "webpack-production": "NODE_ENV=production gulp rev & webpack --color --config webpack.production.config.js",
```
simonista: why get rid of NODE_ENV=production here?

package.json
```diff
@@ -152,7 +152,7 @@
...
+    "webpack-production": "gulp rev & node --max_old_space_size=4096 node_modules/webpack/bin/webpack.js --progress --color --config webpack.production.config.js",
```
simonista: can we not add `--progress`, it makes the output of our build really annoying to read

simonista:
Sorry for the delay on this one @grahamb, turns out September was busy :)

simonista:
Nevermind, we'll just make the changes internally and get it reviewed :)

simonista:
Merged in 5c9ae8f6aa117bbfec297854fb6e3c7cf5768f2a

PR #1107, opened 2017-09-07
Switched rate_limit_exceeded header from 403 to 429

MasterChief-John-117:
Changed rate_limt_exceeded code from 403 to 429

Current response on rate_limit_exceeded is a 403 (Forbidden). IETF (Internet Engineering Task Force) documentation defines a [429 (Too Many Requests)](https://tools.ietf.org/html/rfc6585#section-4) as the proper response code. Code and documentation has been changed to reflect that.

### DIFF
--- a/app/middleware/request_throttle.rb
+++ b/app/middleware/request_throttle.rb
@@ -173,7 +173,7 @@ class RequestThrottle
   end
 
   def rate_limit_exceeded
-    [ 403,
+    [ 429,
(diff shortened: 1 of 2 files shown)


tierra:
Hi @MasterChief-John-117, thanks for your contribution.

Unfortunately, since we've been using this code for a long time (4 years supposedly), changing it would be a backwards incompatible change to the API. We can't actually make this change without bumping the API to a new version (which also can't be done currently).

PR #1117, opened 2017-09-13
Ignore yarn-error.log files

grahamb:
yarn-error.log files should be ignored, just like npm-debug.log files already are.


simonista:
Merged in a6a1801ece414f0cc287807f896d8f9d6e86a3b5

PR #1136, opened 2017-10-16
Fix docker pg build

maths22:
Probably broken by the release of PG 10, but it could be something else.

test plan:
- Check out the commit
- Try `dc build --no-cache postgres`
- It should successfully build


omarkhan:
Thanks for this @maths22! We actually hit this issue this morning. Validating your fix now...

Also, we miss you Jacob!

maths22:
Oh hey Omar!  I miss you guys too.  I'm looking forward to being back next year.

tierra:
Nice find, thanks Jacob!

PR #1137, opened 2017-10-17
Improve score details and grade statistics

maths22:
Adds median and quartile calculations, and uses them to generate a
proper box and whiskers plot. This updates the plot to match the
existing documentation more closely (suggesting that this matches
what a desired product feature).

Example of the score details:
image

test plan:
- Grade an assignment for at least 5 students
- Verify the score details show median and quartile numbers
- Verify the box plot matches those numbers and looks reasonable
- The performance of this code should also be verified for a large
  class size


omarkhan:
Hi Jacob! Don't you have classes or something? 😉 

I see you closed this. Let me know if you want me to find a reviewer for it.

PR #1170, opened 2017-12-01
This change makes the default_view setting available to pandarus

eriko:
While the 'default_view' setting is json api and and is settable via the api it is not in the documentation completely.  This adds it to the documentation so that pandarus can read it while generating the gem.

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -656,7 +656,7 @@ class CoursesController < ApplicationController
   #
   # @argument enroll_me [Boolean]
   #   Set to true to enroll the current user as the teacher.
-   #
+  #
   # @argument course[default_view]  [String, "feed"|"wiki"|"modules"|"syllabus"|"assignments"]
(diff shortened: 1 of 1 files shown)


simonista:
Thanks @eriko, I'll get this run through our test suite.

eriko:
I think that early on there was an issue with arrays in method signatures so it was left out.

app/controllers/courses_controller.rb
```diff
@@ -656,6 +656,15 @@ def user_index
   #
   # @argument enroll_me [Boolean]
   #   Set to true to enroll the current user as the teacher.
+   #
```
simonista: the extra space in front of this `#` seems to be confusing the parser :( _sigh_, would you mind fixing it?

bootstraponline:
Merged.

eriko:
I am sorry I missed your reply on the pull request.  Thanks for fixing my mistake.

PR #1171, opened 2017-12-04
Update ColorPicker.js

atiakabir:
To chang the predefined colours of ColorPicker file to RMIT colours


Sharla:
Unsubscribe 

--
Sharla Sava, PhD

PR #1185, opened 2017-12-20
Reverting due to frontend bug

captainhurst:


simonista:
I think maybe you meant to request this against your own master branch? I'm going to close it here; huge change with no description = seems like an accident.

PR #1224, opened 2018-02-02
Added ID into CustomColumn model so API documentation is accurate.

justin0022:
Left description for teacher_notes blank, as I don't know exactly what it does.

### DIFF
--- a/app/controllers/custom_gradebook_columns_api_controller.rb
+++ b/app/controllers/custom_gradebook_columns_api_controller.rb
@@ -26,6 +26,16 @@
 #       "id": "CustomColumn",
 #       "description": "",
 #       "properties": {
+#         "id": {
+#           "description": "The ID of the custom gradebook column",
+#           "example": "2",
+#           "type": "integer"
+#         },
+#         "teacher_notes": {
+#           "description": "",
+#           "example": false,
+#           "type": "boolean"
+#         },
 #         "title": {
 #           "description": "header text",
 #           "example": "Stuff",
--- a/app/controllers/custom_gradebook_columns_api_controller.rb
+++ b/app/controllers/custom_gradebook_columns_api_controller.rb
@@ -28,7 +28,7 @@
 #       "properties": {
 #         "id": {
 #           "description": "The ID of the custom gradebook column",
-#           "example": "2",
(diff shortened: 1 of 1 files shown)


spencerolson:
Hey @justin0022 , thanks for your contribution! The "teacher_notes" column is the column that will be shown/hidden when a teacher selects to show/hide notes in the Gradebook (other custom columns will not be affected by this action). Maybe you can change the description to something like:

"When true, this column's visibility will be toggled in the Gradebook when a user selects to show or hide notes"

Once you've updated the description, I'll make sure this gets in front of the appropriate team for code review.

Thanks,
Spencer

justin0022:
Hi @spencerolson, thanks!
I've updated the description now!

spencerolson:
Hi @justin0022 , thanks for making that change :D I'll get this in front of devs for code review and keep you updated.

justin0022:
Thanks @spencerolson!

app/controllers/custom_gradebook_columns_api_controller.rb
```diff
@@ -26,6 +26,16 @@
 #       "id": "CustomColumn",
 #       "description": "",
 #       "properties": {
+#         "id": {
+#           "description": "The ID of the custom gradebook column",
+#           "example": "2",
```
spencerolson: Hi @justin0022 , can you push up a new commit with this example value ("2") unquoted? Since it's an integer, we want to drop the quotes here in the example.

spencerolson:
Hi @justin0022 , I left a comment on your commit. Can we unquote the example value ("2")? Once that's done, I think this will be ready to merge :D 

Thanks,
Spencer

justin0022:
Good point. I've made the change now.

spencerolson:
Wow! You're speedy. Thanks for doing that so quickly. Running it through tests again and then I think we'll be good-to-go.

spencerolson:
Hi @justin0022 , your commit has been merged to the `master` branch and can be found here: https://github.com/instructure/canvas-lms/commit/bccdadb1ea2623fdbbbf5f08831ee1937e3a561a .Thanks again for your contribution!

PR #1285, opened 2018-04-27
integration_id validation added.

sirruf:
If we try to create a user with the same id as an existing or remote user, we get the following error:

PG::UniqueViolation: ERROR:  duplicate key value violates unique constraint "index_pseudonyms_on_integration_id"
DETAIL:  Key (integration_id, account_id)=(0001, 1) already exists.

Adding this validation fixes this problem.

### DIFF
--- a/app/models/pseudonym.rb
+++ b/app/models/pseudonym.rb
@@ -44,7 +44,7 @@ class Pseudonym < ActiveRecord::Base
   before_destroy :retire_channels
 
   before_save :set_password_changed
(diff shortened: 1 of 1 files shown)


sirruf:
I've updated both methods.

omarkhan:
This was cherry-picked to master in 30c2b1b, with a couple of minor changes based on review feedback and lint warnings. Thanks for the contribution @sirruf!

PR #1288, opened 2018-05-01
canvas.docker is a development only hostname

buckett:
When the documentation is published to http://api.instructure.com/ the link to http://canvas.docker isn't valid, making it host relative means it works in both places.


omarkhan:
Thanks @buckett. This was cherry-picked to master as 8e78384

Issue #1268, opened 2018-03-30
Discussion: Default branch changing to master


bootstraponline:
To improve the pull request experience on Canvas LMS, the default branch will be changing to master. This allows us to update the documentation in the repo faster compared to stable. By default, pull requests will target the correct branch.

If you rely on stable being the default branch, you'll want to update your scripts to checkout stable explicitly.

- `git clone https://github.com/instructure/canvas-lms.git`
- `cd canvas-lms`
- `git checkout stable`

If there are any problems with this change happening, comment and let us know.

grahamb:
SFU has no problems with this.

christopher-b:
This won't cause any disruption for us at OCADU. Thanks for the heads-up.

bootstraponline:
The default branch has been updated to master.

Another benefit of this change is that GitHub issues will be closed automatically when fixes land to master. Previously issues didn't auto close because stable was the default branch.

PR #1294, opened 2018-05-14
Add a pull request template

tbodt:
Looking through the list of recent pull requests, a surprising number of them are attempting to merge thousands of commits. I thought a pull request template telling people to check for that might help.


tbodt:
...I based this on the wrong branch 🤦‍♂️

tierra:
This typically happened because the default branch was set to `stable` for years, right up until last month with #1268. It's unlikely to continue happening now.

PR #1300, opened 2018-05-19
Docker package renamed

sau226:
Docker's package has been renamed to docker-ce. This commit reflects that rename and unbreaks it on Ubuntu 16.04 LTS

### DIFF
--- a/script/docker_dev_setup.sh
+++ b/script/docker_dev_setup.sh
@@ -41,7 +41,7 @@ if [[ $OS == 'Darwin' ]]; then
   dependencies='docker docker-machine docker-compose dinghy'
 elif [[ $OS == 'Linux' ]]; then
   install='sudo apt-get update && sudo apt-get install -y'
-  dependencies='docker docker-compose'
+  dependencies='docker-ce docker-compose'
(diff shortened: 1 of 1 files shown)


omarkhan:
Thanks for this @sau226. I don't see a `docker-ce` package in the Ubuntu 16.04 repos, but it looks like the `docker` package is actually some kind of system tray thing. Looks like it should be `docker.io` instead: https://packages.ubuntu.com/xenial/docker.io. I imagine it works because `docker.io` is a dependency of `docker-compose`.

tierra:
@sau226 I was wondering what you think about having this script setup the official Docker repository for more up-to-date packages?

sau226:
I'm unsure on how I could implement such a feature. You may merge this now or use your "edit from maintainer" access to try to implement the change in my branch and therefore in this request

bootstraponline:
See https://docs.docker.com/install/linux/docker-ce/ubuntu/#install-using-the-repository

bootstraponline:
Thanks for the pull request. If you have more time in the future, feel free to resubmit.

PR #1326, opened 2018-07-24
BBB: Fix for issue with recording duration always shown as '1 minute'

jfederico:
### Summary:
Recording duration is not correctly displayed when statistics are enabled. This is because the condition that filters the playback format used as a reference is incorrect.

### Test Plan:
1. Configure the BBB server to be used by the Plugin. Make sure the recordings are enabled.
2. Go to the Course/Conferences
3. Create a Conference. Enable recordings. Save
4. Start the conference. This will send you to BBB. Start the recording and keep it open for more than 2 minutes.
5. Logout from BBB ending the meeting.
6. Go back to Canvas conferences
7. Wait for a few minutes until the recording is processed. You can refresh the page to see if the result has come trough.
8. Once the recording is shown, expand the recording and verify the duration. It should say "Duration: 2 minutes" (or 3 minutes if you recorded more than 2 mins 30 secs)

### Expected behavior:
As the seconds are rounded, the normal behaviour would be to show:

"1 minute"  when the session lasted from 0 seconds to 1 minute with 30 seconds.
"n minutes" whhen the meeting lasted from 1 minute 31 secs to 59 minutes 30 seconds
"m hours n minutes" when the meeting lasted more than 59 minutes and 31 seconds

### Development ID:
https://github.com/blindsidenetworks/canvas-lms/issues/30

### DIFF
--- a/app/models/big_blue_button_conference.rb
+++ b/app/models/big_blue_button_conference.rb
@@ -238,9 +238,10 @@ class BigBlueButtonConference < WebConference
   end
 
   def filter_duration(recording_formats)
-    # As not all the formats are the actual recording, identify the one that has :length
+    # This is a filter to take the duration from any of the playback formats that include a value in length.
+    # As not all the formats are the actual recording, identify the first one that has :length <> nil
     recording_formats.each do |recording_format|
-      return recording_format[:length].to_i if recording_format.key?(:length)
+      return recording_format[:length].to_i if recording_format[:length].present?
     end
   end
 end
--- a/spec/fixtures/files/conferences/big_blue_button_get_recordings_two.json
+++ b/spec/fixtures/files/conferences/big_blue_button_get_recordings_two.json
@@ -23,7 +23,7 @@
         {
           "type": "presentation",
           "url": "https://bbb.blah.com/instructure/18974fe54920ac60ba913e34f49e4a9dabfeea2c-1513031612456/presentation/",
-          "length": "0",
+          "length": "2",
           "preview": {
             "images":
             [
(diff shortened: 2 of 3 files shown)


venknatarajan:
A few comments (for some reason I'm not able to do inline comments on the code, so bear with me):

* You'll need a test plan for your commit -- put in a description of how someone can test that this works as intended. 
* Also add in some specs around this.
* `# As not all the formats are the actual recording, identify`: This is not true anymore, so remove this comment while you're touching this anyway
* `return recording_format[:length].to_i if recording_format[:length]`: "Will this return out of the each or out of the function?"

jfederico:
Hi @venknatarajan, thanks for responding this quick. As I was pointing out in slack this PR is really upon request of the Instructure support team. We can follow-up on slack if you want.

>You'll need a test plan for your commit -- put in a description of how someone can test that this works as intended.
Yeah, I had some steps in the description before, but they were defined as "Steps to reproduce", anyways, the Test case scenario is there now.

>Also add in some specs around this.
This is really a very simple bug/patch. Since the issue is in the model I added a test for making sure the value is never 0 when presentation or video formats have some value grater than 0. It is based on the fixture. 
I also updated the fixture and removed redundancy from the spec.

> # As not all the formats are the actual recording, identify: This is not true anymore, so remove this comment while you're touching this anyway
This I don;t know where you found it. But it is fair to say that I try to change the less possible lines to make life easier for the guy in engineering. We usually go through a deeper round of updates there.

>return recording_format[:length].to_i if recording_format[:length]: "Will this return out of the each or out of the function?"
This is a filter to take the duration from any of the playback formats that include a value in length. Some formats like statistics, as they are not recordings, don't include any value (they are equal to nil). So, yes, the intention is that the loop breaks and the methods ends when the value is found. If there was only a format without length, the response would be nil

venknatarajan:
This has been code-review approved, now going to QA

tierra:
This is merged in faf9eb42d28c674145209ff9365759877cbf26c6

PR #1345, opened 2018-09-14
Adding workflow state to syllabus assignment rows, and course summary h2

robert-carroll:
Adding attributes for specificity on the course syllabus
Attributes can be styled or interacted with, to delineate un/published assignments in the syllabus


robert-carroll:
closing to resubmit without cla issues.

PR #1346, opened 2018-09-18
add data-attributes for syllabus assignments summary

robert-carroll:
...and syllabus assignment TR workflow state.

I would like to add an attribute to assignment rows with the workflow-state of the assignment. There currently seems to be no distinction between published/unpublished assignments. Instructors are unable to determine the difference. Adding the `data-workflow-state` attribute would allow styling at a minimum. see https://community.canvaslms.com/thread/8283#comment-117711

Tested with CSS after updating and running source locally.
```css
table#syllabus tr[data-workflow-state="unpublished"] td a {
    font-weight: bold;
}
```

### DIFF
--- a/app/coffeescripts/views/courses/SyllabusView.coffee
+++ b/app/coffeescripts/views/courses/SyllabusView.coffee
@@ -156,6 +156,7 @@ define [
           'last': true
           'override': override
           'json': json
+          'workflow_state': json['workflow_state']
 
         lastDate['events'].push lastEvent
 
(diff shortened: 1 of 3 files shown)


simonista:
The h2 data element seems strange. I've been chatting with @robert-carroll on IRC about alternatives, he's going to try a couple.

robert-carroll:
I've decided to table the attribute on the H2, I'll find an alternative solution.
After installing and running Canvas locally, I found that I needed to update the data source for the view in SyllabusView.coffee to get workflow_state to the template.

Issue #1341, opened 2018-09-10
Rubric scopes are missing


Boudewijn26:
### Summary:
It's impossible to add the `url:POST|/v1/api/courses/:course_id/rubrics` and `url:PUT|/v1/api/courses/:course_id/rubrics/:id` scopes to a developer key

### Steps to reproduce:

 1. Make sure the 'Developer key management and scoping' feature option is enable
 1. Create or edit a developer key
 2. Enable enforce scopes
 3. Try to add `url:POST|/v1/api/courses/:course_id/rubrics` and `url:PUT|/v1/api/courses/:course_id/rubrics/:id`

### Expected behavior:
The `url:POST|/v1/api/courses/:course_id/rubrics` and `url:PUT|/v1/api/courses/:course_id/rubrics/:id` are available under Rubrics.

### Actual behavior:
The aforementioned scopes are missing.
![image](https://user-images.githubusercontent.com/2896917/45297485-d488d000-b505-11e8-91a9-e87f9d3d626f.png)

### Additional notes:
While these endpoint [aren't documented](https://canvas.instructure.com/doc/api/rubrics.html) they provide very useful functionality for tools that want to send extended grading information.

Boudewijn26:
@ccutrer Would it be possible to give something of an indication here?

ccutrer:
Are you trying to add /api/v1, or /v1/api?

Boudewijn26:
My apologies, /api/v1/ indeed

PR #1349, opened 2018-09-20
add missing rubric scopes

Boudewijn26:
The missing @API tag caused the RubricsController to be missing from the yard `options[:objects]` which are [transformed](https://github.com/instructure/canvas-lms/blob/master/doc/api/fulldoc/html/setup.rb#L159) into the resources which are handed to the ApiScopeMappingWriter.

Since the ApiScopeMappingWriter didn't look at the actions provided in the RubricsController, the scopes weren't written to lib/api_scope_mapper.rb which means that they can't be enabled in
the developer keys.

closes gh-1341

Test Plan:
    - Run either `rails doc:api` or `rails canvas:compile_assets`
    - Make sure the feature option  Developer key management and scoping is enabled in your top level account
    - Go the the admin of your top level account
    - Create a new developer key
    - Enable enforce scopes
    - See if the url:POST|/api/v1/courses/:course_id/rubrics and url:PUT|/api/v1/courses/:course_id/rubrics/:id scopes are present under the rubrics expander

### DIFF
--- a/app/controllers/rubrics_controller.rb
+++ b/app/controllers/rubrics_controller.rb
@@ -16,6 +16,7 @@
 # with this program. If not, see <http://www.gnu.org/licenses/>.
 #
 
+# @API Rubrics
 class RubricsController < ApplicationController
   before_action :require_context
   before_action { |c| c.active_tab = "rubrics" }


simonista:
Hey @Boudewijn26, thanks for the PR. I was reviewing it, and noticed that the api documentation is wrong in https://github.com/instructure/canvas-lms/blob/master/app/controllers/rubrics_controller.rb (not what you added, but what was already there). It claims that create and delete return a rubric object, but they don't, they return non-standard json that includes both the rubric and the assessment as top level keys.

I need to think through what the right set of steps are here, and whether we need to correct the return value before making the api documentation publicly visible. I'll get back to you about that soon!

Boudewijn26:
Hi @simonista,

Any updates so far?

simonista:
I think we're good with this. I'll pull it in for final review to be merged.

simonista:
merged in 6c25a6cbcc704f4ce9807b3e5ebe8fad76aaa687

simonista:
FYI, I also added some extra explanation about the wonky return values in f3ae09158c714c9ad8f6f2db81801767b7eb85eb

PR #1366, opened 2018-10-19
Support LTI Request for Page Information

slaughter550:
# Description

@notebowl is attempting to make the LTI interface more seamless. One of the things that would help a lot is to know the total width and height of the window along with the iFrame's offset. This will help us best display & scale our own elements inside of the iFrame.

## Usage

From the iFrame:
```
// Setup listener to grab response
window.addEventListener('message', function(e) {
    var message = JSON.parse(e.data);
    if(message.subject === 'lti.fetchWindowSize') {
        console.log(message.height);
        console.log(message.width);
        console.log(message.offset);
    }
});

// Ask parent window for the information to be sent to the page event listener
message = {'subject': 'lti.fetchWindowSize'};
window.parent.postMessage(JSON.stringify(message), '*');
```

### DIFF
--- a/public/javascripts/lti/messages.js
+++ b/public/javascripts/lti/messages.js
@@ -74,6 +74,18 @@ export function ltiMessageHandler(e) {
       }
       break;
 
+      case 'lti.fetchWindowSize':
+          const iframe = findDomForWindow(e.source);
+          if (iframe) {
+              message.height = window.innerHeight;
+              message.width = window.innerWidth;
+              message.offset = $('.tool_content_wrapper').offset();
+              var strMessage = JSON.stringify(message);
+
+              iframe.contentWindow.postMessage(strMessage, '*');
+          }
+        break;
+
       case 'lti.showModuleNavigation':
         if(message.show === true || message.show === false){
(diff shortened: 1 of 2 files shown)


jbasdf:
Canvas let's an LTI tool pass up changes in content size using lti.resize. I think it makes sense to let the information flow down into the tool as well.

cmfurse:
@slaughter550 : we have a few lint issues before merge if you could fix:

public/javascripts/lti/messages.js
Line 78:
[WARN] [eslint] Unexpected lexical declaration in case block

Line 83:
[WARN] [eslint] All 'var' declarations must be at the top of the function scope

[WARN] [eslint] Unexpected var, use let or const instead

Thanks!

slaughter550:
@cmfurse I believe it's fixed! Thanks!

cmfurse:
@slaughter550 

sorry for the delay...here is one more:

public/javascripts/lti/messages.js
Line 83:
[WARN] [eslint] 'strMessage' is never reassigned. Use 'const' instead

slaughter550:
@cmfurse fixed! Thanks!

westonkd:
Merged: 69494f4

PR #1379, opened 2018-11-07
Updating Live Events (Canvas Raw) Documentation

robert-carroll:
Looking for some collaborative editing.

I've added and updated some items at the top. Alphabetized the events. Added events, and fields that I've collected from messages. I've also added data types for the fields I've seen, based on the data I have received. Some of that might need to be double checked. For instance, I was unaware until yesterday that `sis_source_id` and `sis_batch` ID can be strings, I have corrected that. I have tried to normalize the data types to the Canvas Data schema. I understand 'String' is used in the metadata table and that a lot of this data may be strings in other instances. I'm open to suggestions, but figured this was a good starting point.

There are also some fields I need internal input to define. I've left them blank, find them by looking for '_eventfielddescription_'.

Does not include the _Quizzes.Next_ events. Considerations for adding those in a separate document, or adding here?

test plan:
- consumed 5M+ messages in an auto generating test environment creating tables for events and adding columns when if new fields are seen. I believe I have everything that is currently sent.
- created a DDL with specific data types and then explicitly set those fields and data types in code
- continued importing and fixing datatype and length issues when found
- generated markdown with data types from database

### DIFF
--- a/doc/api/live_events.md
+++ b/doc/api/live_events.md
@@ -16,41 +16,53 @@ The currently supported queue is <a href="https://aws.amazon.com/sqs/">Amazon SQ
 Live Events is currently an invite-only, experimental feature.
 
 
-### Event Format
+Additional Documentation can be viewed at
+[Live Events Services Table of Contents](https://community.canvaslms.com/docs/DOC-15740-live-events-services-table-of-contents)
+
+
+### Live Events - Canvas Raw Format
+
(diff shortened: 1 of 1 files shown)


simonista:
Thanks @robert-carroll, I'll try to recruit some people to help with this. Keep us honest if this sits for too long.

PR #1380, opened 2018-11-09
Add `created_at` to group categories API

opes:
Test Plan:
* Group category APIs should return their respective
  object's `created_at` field


simonista:
Hey @opes, sorry for the slow response. Looks good to me, I'm just pulling it into gerrit and running tests.

PR #1405, opened 2019-01-17
Play 844 fix course color picker

AndyRosenbergStrongMind:


claydiffrient:
@AndyRosenbergStrongMind We've noticed these large PRs come through for a while now.  Is there anything that can be done to prevent these from happening?  They just sorta clutter everything up on this end.

PR #1407, opened 2019-01-23
Prevent page error when Dashcard position contains invalid characters.

nijssen:
When set_dashboard_positions is given a string instead of an integer, they are happily stored in the user record.
Later when the dashboard is rendered, Ruby tries to sort the dashcards by :position and this means that a string is compared against an int, crashing the application with "comparison of String with 2 failed" or something similar.

addresses Canvas Case #03910104

 Test Plan:
   - Call /api/v1/users/self/dashboard_positions with something that looks like an integer but actually isn't, such as {"dashboard_positions":{"course_508213":"0  ","course_598941":2,"course_598980":1,"course_661886":3}}
   - Ensure that there is no page error on next dashboard render.

### DIFF
--- a/app/helpers/dashboard_helper.rb
+++ b/app/helpers/dashboard_helper.rb
@@ -155,7 +155,7 @@ module DashboardHelper
       presenter.to_h
     end
 
-    mapped.sort_by {|h| h[:position] || ::CanvasSort::Last}
+    mapped.sort_by {|h| Integer(h[:position] || ::CanvasSort::Last)}
   end
 
 end


nijssen:
any update on this?

SecurityInsanity:
Hey @nijssen ,

Looks like this happened before our auto assigning bot. We'll get this assigned. Sorry for the delay!

sdb1228:
Hey @nijssen Sorry it has taken so long to get a review on this thing, but I am a little confused how we are even getting string positions in? it looks like we explicitly check if its an integer on save. 
 https://github.com/instructure/canvas-lms/blob/master/app/controllers/users_controller.rb#L1789

I think this is fine to prevent the crash the opposite way but I would be interested to know how we are even saving strings when the check above is happening. 

Could we also get a unit test around this by chance?

nijssen:
The reason I am able to store strings is because those strings are integer-like. That is, `Integer("1") == Integer("1           ") == 1` (notice the spaces). Therefore, the `Integer(val)` check in the code you linked passes if given a string. On line 1795, the supplied params are merged into what is stored in the database column (which is serialized, preserving the types).
No problem just yet.

However, when this is retrieved and [sorted](https://github.com/instructure/canvas-lms/blob/dcb3975dc1c01781d36c045412d4d7ee55221927/app/helpers/dashboard_helper.rb#L158), Ruby complains about the mismatched types, because some of the positions were stored as integers (as expected) while some are stored as strings. 

As for a unit test, I'm not sure how to make one as I am not a Ruby programmer. My fix is based on intuition and Googling :-)

sdb1228:
Ah I see I figured we were storing that position in the db after parsing but glanced too fast.  Looks like we are storing their input as is https://github.com/instructure/canvas-lms/blob/master/app/controllers/users_controller.rb#L1795.  I think this is fine for now.  May be worth stripping the whitespace before merging but it might be overkill.  As for a spec I would copy one found here https://github.com/instructure/canvas-lms/blob/master/spec/helpers/dashboard_helper_spec.rb#L76

claydiffrient:
@nijssen - I went ahead and took the liberty of writing a spec in for this change locally and tested it out.

The spec is as follows placed at line 100 of `dashboard_helper_spec.rb`
```ruby
it "handles sorting even when positions are strings" do
  course1 = @account.courses.create!
  course2 = @account.courses.create!
  course3 = @account.courses.create!
  user = user_model
  course1.enroll_student(user)
  course2.enroll_student(user)
  course3.enroll_student(user)
  courses = [course1, course2, course3]
  user.dashboard_positions[course1.asset_string] = 3
  user.dashboard_positions[course2.asset_string] = "2"
  user.dashboard_positions[course3.asset_string] = 1
  user.save!
  @current_user = user
  mapped_courses = map_courses_for_menu(courses)
  expect(mapped_courses.map {|h| h[:id]}).to eq [course3.id, course2.id, course1.id]
end
```

Feel free to add that to your commit.

Also, it appears that something is off here because this commit as is causes other test failures notably the following:
```
bin/rspec ./spec/controllers/users_controller_spec.rb:2118
bin/rspec ./spec/controllers/users_controller_spec.rb:2125
bin/rspec ./spec/controllers/users_controller_spec.rb:2133
```
Could you take a look and see if there's an alternative change you can make to accomplish the same?  Or perhaps a way we can change the specs to accommodate the new change?

simonista:
Closed in favor of #1561

PR #1412, opened 2019-01-31
853 fix enrollment fetcher

chrisyoung:


SecurityInsanity:
Hey @chrisyoung ,

We've noticed these large PRs come through for a while now. Is there anything that can be done to prevent these from happening? They just sorta clutter everything up on this end.

Issue #1443, opened 2019-05-10
iFrame not loading css files MIME Type problem


diegotejadav:
### Summary:
When trying to embed a content hosted in the course files section with an iframe the .css, .jpg and .js files aren't getting loaded and Chrome's inspector shows: "Refused to apply style from 'http://192.168.1.13/courses/2/files/11/css/bootstrap.min.css' because its MIME type ('text/html') is not a supported stylesheet MIME type, and strict MIME checking is enabled."

![mime_type_error](https://user-images.githubusercontent.com/3412546/57561196-f87c0a00-734f-11e9-86fe-2bef2076d416.png)

### Steps to reproduce:

 1. Create a simple html page with an image and some styles (css)
 2. Then upload the files (html, css, js) in any course in the **files** section
 3. Create a new page for the course
 4. Select the .html file, then the editor will show the relative path it will show an href like this:
```html
<a target="" title="index.html" href="/courses/2/files/11/download?wrap=1" class="instructure_file_link ">index.html</a>
```
 5. Then get the relative path `/courses/2/files/11/download` and make an iframe with the absolute path:
```html
<iframe src="http://192.168.1.13/courses/2/files/11/download" width="100%" height="700" data-api-endpoint="http://192.168.1.13/api/v1/courses/12/files/1356" data-api-returntype="File"></iframe>
```
 6. Save the page.
 
### Expected behavior:
After the page is saved the html page should be shown with full css styles and so on.

### Actual behavior:
Just the html is shown and the .css files are getting blocked and Chrome's and Mozilla's inspector show the error message:
_Refused to apply style from 'http://192.168.1.13/courses/2/files/11/css/bootstrap.min.css' because its MIME type ('text/html') is not a supported stylesheet MIME type, and strict MIME checking is enabled._
### Additional notes:
This is the preview of the .html file, .css files are getting blocked as shown before
![image (4)](https://user-images.githubusercontent.com/3412546/57561486-902e2800-7351-11e9-873a-b41f2c40b1ba.png)
These are the uploaded files in the course:
![image (5)](https://user-images.githubusercontent.com/3412546/57561513-af2cba00-7351-11e9-87b5-260d4e73e739.png)
This is the page preview, with the same blocking message as shown before.
![image (6)](https://user-images.githubusercontent.com/3412546/57561567-016ddb00-7352-11e9-96f8-c0365c7234ad.png)

jorgemorahenao:
Same problem here!

[...]

Issue #1413, opened 2019-02-04
image not loading


Bhalani:
### Summary:
When I load the preview of images, it's not loading because of `Resource interpreted as Document but transferred with MIME type image/jpeg:`

But When I manually add extension `.image` at end of image src  it is loading the image.

### Steps to reproduce:

 1. Upload a image in profile pocture
 2. click on image to preview that
 3.

### Actual behavior:
I don't know whether it should load the image from a action without mimetype like .jpg, .png, .image(any image)

diegotejadav:
@Bhalani, have any news on this? Looks like related to my problem at #1443

diegotejadav:
I don't really know if that's a misconfiguration on the Apache server

Bhalani:
@diegotejadav 
So I found the code file where that preview image was setup by url
I just edit that and setup `.image` MIME type at end of url.
File I have edited is var/canvas/app/views/file_previews/img_preview.html.erb

PR #1449, opened 2019-05-19
- Solve issue of loading image without extension

Bhalani:
Trying to add solution for these issues.
https://github.com/instructure/canvas-lms/issues/1413
https://github.com/instructure/canvas-lms/issues/1443


simonista:
Hi @Bhalani, thanks for the sending us this PR. Unfortunately we won't be able to accept it as is. We can't just append `.image` to every image preview url, that won't work.

I can't reproduce the issue you're having in #1413, so maybe we can discuss more there and try to pin down the problem, but it sounds like it might have something to do with your setup.

PR #1459, opened 2019-05-29
Add colon to license label in course settings page

Gao-Jun:
In course settings page, all filed name is end with a colon except license.

This patch add colon to license field to unify the style in the page.

Before:
before

After:
after

### DIFF
--- a/app/views/courses/settings.html.erb
+++ b/app/views/courses/settings.html.erb
@@ -334,8 +334,8 @@
         </td>
       </tr>
       <tr>
-        <td class="form-label"><%= f.label :license, :en => "License" %></td>
-        <td colspan="3">
+        <td class="form-label"><label for="license"><%= before_label('license', %{License}) %></label></td>
+        <td colspan="3" id="license">
           <% if can_manage %>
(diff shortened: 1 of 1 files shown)


app/views/courses/settings.html.erb
```diff
@@ -334,8 +334,8 @@
         </td>
       </tr>
       <tr>
-        <td class="form-label"><%= f.label :license, :en => "License" %></td>
-        <td colspan="3">
+        <td class="form-label"><label for="license"><%= before_label('license', %{License}) %></label></td>
+        <td colspan="3" id="license">
```
claydiffrient: Technically this change isn't needed because there's already an id on the select element (which is where the label for should point)

app/views/courses/settings.html.erb
```diff
@@ -334,8 +334,8 @@
         </td>
       </tr>
       <tr>
-        <td class="form-label"><%= f.label :license, :en => "License" %></td>
-        <td colspan="3">
+        <td class="form-label"><label for="license"><%= before_label('license', %{License}) %></label></td>
```
claydiffrient: This for should be `course_license` so that the label associates with the proper control.

claydiffrient:
Merged in as d9c40c12778437bf536b168e0fd3085d196c0ad4

Gao-Jun:
That's fine, thank you.

PR #1463, opened 2019-06-03
Fix path to browsers.yml relative to browser_support.rb

shawniverson:
Problem
The relative depth to browsers.yml is one level too high.

Solution
Set relative depth to one level up instead of two levels.


shawniverson:
My instance reverted to a new relative path.   /var/canvas/lib/config/browsers.yml caused by No such file or directory @ rb_sysopen

Closing, invalid.

Issue #827, opened 2016-05-20
bundle install : can install thrift


gabrieltong:
ruby version 2.1.5 , gem version 1.10.6

```
Gem::Ext::BuildError: ERROR: Failed to build gem native extension.

    /Users/shangliang/.rvm/rubies/ruby-2.1.5/bin/ruby -r ./siteconf20160520-20541-1undjkx.rb extconf.rb
checking for strlcpy() in string.h... yes
creating Makefile

make "DESTDIR=" clean

make "DESTDIR="
compiling binary_protocol_accelerated.c
compiling compact_protocol.c
compiling memory_buffer.c
compiling protocol.c
compiling strlcpy.c
In file included from strlcpy.c:20:
./strlcpy.h:28:15: error: expected parameter declarator
extern size_t strlcpy(char *, const char *, size_t);
              ^
/usr/include/secure/_string.h:105:44: note: expanded from macro 'strlcpy'
  __builtin___strlcpy_chk (dest, src, len, __darwin_obsz (dest))
                                           ^
/usr/include/secure/_common.h:39:62: note: expanded from macro '__darwin_obsz'
#define __darwin_obsz(object) __builtin_object_size (object, _USE_FORTIFY_LEVEL > 1 ? 1 : 0)
                                                             ^
/usr/include/secure/_common.h:30:32: note: expanded from macro '_USE_FORTIFY_LEVEL'
#    define _USE_FORTIFY_LEVEL 2
                               ^
In file included from strlcpy.c:20:
./strlcpy.h:28:15: error: expected ')'
/usr/include/secure/_string.h:105:44: note: expanded from macro 'strlcpy'
  __builtin___strlcpy_chk (dest, src, len, __darwin_obsz (dest))
                                           ^
/usr/include/secure/_common.h:39:62: note: expanded from macro '__darwin_obsz'
#define __darwin_obsz(object) __builtin_object_size (object, _USE_FORTIFY_LEVEL > 1 ? 1 : 0)
                                                             ^
/usr/include/secure/_common.h:30:32: note: expanded from macro '_USE_FORTIFY_LEVEL'
#    define _USE_FORTIFY_LEVEL 2
                               ^
./strlcpy.h:28:15: note: to match this '('
/usr/include/secure/_string.h:105:44: note: expanded from macro 'strlcpy'
  __builtin___strlcpy_chk (dest, src, len, __darwin_obsz (dest))
                                           ^
[...]

PR #1466, opened 2019-06-11
Include missing scopes in error message

Boudewijn26:
Since there are quite a lot of possible scopes, it can be difficult to make sure every requested scope is enabled. The error message when a requested scope is missing, doesn't include the scope, which makes this even more difficult. This PR adds a list of missing scopes to the error message.

Test Plan:
  - Set up a developer key with scopes enabled
  - Using a tool like [oauth2-client-shell](https://github.com/neverendingqs/oauth2-client-shell) request access at https://\<canvas-install-url\>/login/oauth2/auth with at least a single scope which is not enabled on the developer key
  - Observe the error message (it should contain the requested scope which is missing, but none of the scopes which are requested, but not enabled).

### DIFF
--- a/app/controllers/oauth2_provider_controller.rb
+++ b/app/controllers/oauth2_provider_controller.rb
@@ -38,7 +38,7 @@ class Oauth2ProviderController < ApplicationController
     raise Canvas::Oauth::RequestError, :invalid_client_id unless provider.has_valid_key?
     raise Canvas::Oauth::RequestError, :invalid_redirect unless provider.has_valid_redirect?
     if provider.key.require_scopes?
-      raise Canvas::Oauth::RequestError, :invalid_scope unless provider.valid_scopes?
+      raise Canvas::Oauth::InvalidScopeError, provider.missing_scopes unless provider.valid_scopes?
     end
 
     session[:oauth2] = provider.session_hash
--- a/lib/canvas/oauth/grant_types/client_credentials.rb
+++ b/lib/canvas/oauth/grant_types/client_credentials.rb
@@ -34,7 +34,7 @@ module Canvas::Oauth
 
       def validate_type
         raise Canvas::Oauth::InvalidRequestError, @provider.error_message unless @provider.valid?
-        raise Canvas::Oauth::RequestError, :invalid_scope unless @provider.valid_scopes?
+        raise Canvas::Oauth::InvalidScopeError, @provider.missing_scopes unless @provider.valid_scopes?
       end
 
       def generate_token
--- a/lib/canvas/oauth/invalid_scope_error.rb
+++ b/lib/canvas/oauth/invalid_scope_error.rb
@@ -1,0 +1,39 @@
+#
+# Copyright (C) 2018 - present Instructure, Inc.
+#
+# This file is part of Canvas.
+#
+# Canvas is free software: you can redistribute it and/or modify it under
+# the terms of the GNU Affero General Public License as published by the Free
+# Software Foundation, version 3 of the License.
+#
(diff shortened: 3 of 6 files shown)


tierra:
Unfortunately, we haven't been able to fully configure CI with GitHub pulls yet (but that is something we're still working on resolving). I did pull your changes into our internal CI system though. The following specs failed with these changes:

```
Rerun 1 Failures
bin/rspec ./spec/controllers/oauth2_provider_controller_spec.rb:45
bin/rspec ./spec/controllers/oauth2_provider_controller_spec.rb:56
```

Additionally, there's a minor rubocop linting issue introduced:

```
lib/canvas/oauth/missing_scopes_error.rb
Line 21:
[INFO] [rubocop] Metrics/LineLength: Line is too long. [200/160]

       super("A requested scope is invalid, unknown, malformed, or exceeds the scope granted by the resource owner. The following scopes were requested, but not granted: #{missing_scopes.to_sentence}")
                                                                                                                                                                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

Boudewijn26:
Thanks tierra, I wasn't able to get Canvas working entirely locally yet (difficulties with building thrift between gcc and clang?), but got enough working to run the tests. They should now pass and include an actual check of the error message and the rubocop linting issue should now be fixed, I hope this follows your code standards.

tierra:
Your thrift issue could be related to #827 btw. It's a somewhat common issue OSX users run into.

tierra:
@ccutrer Do you see anything concerning about these changes? It all looks fine to me.

ccutrer:
Looks fine to me.

tierra:
Sorry this has been sitting for a bit. I've pushed up your latest changes internally, and expect that all CI should pass. This just needs a final code review, then over to QA, where I expect it should go through pretty quickly.

tierra:
Thanks for providing a good test plan by the way, that certainly helps get it through faster.

Boudewijn26:
No need to apologize. I'm happy to see this moving along.

lib/canvas/oauth/missing_scopes_error.rb
```diff
@@ -0,0 +1,39 @@
...
+    def initialize(missing_scopes)
+      super("A requested scope is invalid, unknown, malformed, or exceeds the scope granted by the resource owner. "\
+        "The following scopes were requested, but not granted: #{missing_scopes.to_sentence}")
+    end
+
+    def to_render_data
+      {
+        status: 400,
+        json: {
+          error: :missing_scopes,
```
tierra: One of our devs actually just pointed out that this code change breaks the RFC6749 spec. Note how you've changed the error code from "invalid_scope" to "missing_scopes". It should still be fine that you add the missing scopes to the "error_description" below, but this needs to stay as "invalid_scope".

https://tools.ietf.org/html/rfc6749#section-4.1.2.1

tierra: It might actually be appropriate to even rename this class InvalidScopeError as well.

Boudewijn26:
Thanks for pointing that out, I wasn't that familiar with the spec. I hope this is in your agreement

SecurityInsanity:
Sorry about that! I tagged the wrong Github PR on my merge of another PR. Reopening.

tierra:
Thanks for update @Boudewijn26, I've push your latest changes up. Just another round of reviews, and it should be good again.

tierra:
This is now merged in 2d210624e3bebea09f3e031908969b18b44c2c5a with some extra specs I added to go along with it, thanks @Boudewijn26 !

PR #1467, opened 2019-06-11
Include content migration hook settings in exported LTI config XML

dspencer001:
This PR adds the content migration hooks to the exported XML for LTI tools. These settings were not included, so when a course was copied the import callback would not be called on the copied version of the tool.


brentburg:
Thank you for the pull request. Sorry for the delay. We are currently reviewing this change and I will let you know if we see any issues.

brentburg:
Merged https://github.com/instructure/canvas-lms/commit/c44239d000a6ffbbe7bc119a40af6198822f6695

PR #1479, opened 2019-07-22
Only serve from brotli folder if S3 uploading is enabled

moorecp:
Currently, if the user has a `canvas_cdn.yml` file, Canvas will always append `/br` to the configured CDN host if the client's browser supports brotli. This is an issue if the user has `enabled: false` set in the YAML as Canvas won't build the `/br` directory unless it is uploading assets to S3.

This fix simply checks to see if S3 uploading is also enabled when performing other checks to see if the brotli S3 bucket should be used.

Test Plan:
  - Toggle the `enabled` value in `canvas_cdn.yml` and observe the generated URLs when loading a page.

### DIFF
--- a/lib/canvas/cdn.rb
+++ b/lib/canvas/cdn.rb
@@ -42,7 +42,7 @@ module Canvas
       def add_brotli_to_host_if_supported(request)
         # there is a /br/ folder on the s3 bucket that has evertying we publish,
         # but encoded as brotli instead of gzip
-        "#{config.host}#{'/br' if config.host && supports_brotli?(request)}"
+        "#{config.host}#{'/br' if config.host && config.enabled && supports_brotli?(request)}"
       end
 
       def supports_brotli?(request)


moorecp:
538fdb1 doesn't address the issue. We use the `host` setting for using a CDN, but not the `bucket` setting as we don't have our assets uploaded to S3.  Canvas still tries to serve brotli files even though they don't exist because the only check is if `host` is set and if `should_be_in_bucket?` is `true`. `should_be_in_bucket?` doesn't check for a `bucket` setting; it just checks a path fragment.

moorecp:
I've updated my branch to check for the existence of `config.bucket` in addition to `config.host` (which was already being checked), but I think the closed nature of the PR is keeping it from updating.

ryankshaw:
@moorecp ah ok, so you want a way to all requests to static assets to a cdn host but to a host that doesn't have the /br version of those files?

a couple questions:
* where _are_ you serving them from if not s3? would it make sense to have a build step of your own that makes a copy in `/br` for everything in `/dist`?
* apart from this problem, if you are not using s3, are you setting the right caching headers and stuff? if you look at s3_uploader.rb you can see that it sets some headers that you'll want to have on your assets too

but ya, if it makes sense maybe we could have a config setting in that yml file for `disable_serving_br` that you could set. would that be better than having an implicit check where if both `config.bucket  and config.host` are set that it automatically does the `/br` stuff?

moorecp:
Our assets are being served via CloudFront. They get cached up there upon first request, so they get served once from the web servers. When we first started using Canvas years ago, uploading assets to S3 via canvas_cdn didn't seem to always work as expected, so we just set `enabled` to `false`. I'm guessing there was just a bug years ago that has since been fixed.

If we're talking about having a build step for the `br` versions, wouldn't it just make sense to always build that directory locally when `canvas_cdn` is used irrespective of any S3 specific settings and then just upload it to S3 if there is a bucket?

This is all driven out of a seemingly valid setup being invalidated by the addition of brotli. We could probably just add a bucket to the config and everything would be fine, but that requires more investigation on our end and this bug would still exist for anyone using a CDN without also using S3.

ryankshaw:
got it.
so from your perspective, what do you think makes more sense:
* add the config.bucket check like you did in https://github.com/moorecp/canvas-lms/commit/67257ddc07d9a19ec23e3885b946c7e134f4580c
* or, add a `disable_serving_brotli_assets` setting to the canvas_cdn.yml file?

whichever sounds better to you sounds good to me

moorecp:
Adding `disable_serving_brotli_assets` is more explicit as to what is going on, with the drawback of it still introducing a breaking change. I'm trying to decide if there's a world where a user would want S3 uploads but not want to serve brotli assets. I'm leaning towards no.

I feel like the best option is to modify the build process to build brotli assets every time if a CDN host exists in the YAML file and then just upload them to S3 if a bucket exists. That option is muuuch more involved, though.

Barring that, I'm leaning towards the bucket check (actually, it should be an `enabled?` check since that exists). That is the behavior used in `lib/brandable_css.rb` and the `push_to_s3!` method in `cdn.rb`, so it seems consistent.

moorecp:
@ryankshaw I updated my branch with the change I settled on in my last comment. moorecp@2e35d6a5208638029f6aa5c5e7cd004fd9f507aa Let me know if this works for you.

PR #1480, opened 2019-07-25
Added filter on start_at and end_at for account courses api.

tim-schilling:
Changed AccountsController.courses_api to support filtering courses
that start before a value and end after a value. Courses that have
a null value for the field will still be returned. The term's respective
field will also be checked. Either the course or term field passing
the filter will cause the course to be returned.

Test Plan:
  - Jenkins passes

### DIFF
--- a/app/controllers/accounts_controller.rb
+++ b/app/controllers/accounts_controller.rb
@@ -537,9 +537,24 @@ class AccountsController < ApplicationController
   #   The filter to search by. "course" searches for course names, course codes,
   #   and SIS IDs. "teacher" searches for teacher names
   #
+  # @argument starts_before [Optional, Date]
+  #   If set, only return courses that start before the value (inclusive)
+  #   or their enrollment term starts before the value (inclusive)
+  #   or both the course's start_at and the enrollment term's start_at are set to null.
+  #   The value should be formatted as: yyyy-mm-dd or ISO 8601 YYYY-MM-DDTHH:MM:SSZ.
+  #
+  # @argument ends_after [Optional, Date]
+  #   If set, only return courses that end after the value (inclusive)
+  #   or their enrollment term ends after the value (inclusive)
+  #   or both the course's end_at and the enrollment term's end_at are set to null.
+  #   The value should be formatted as: yyyy-mm-dd or ISO 8601 YYYY-MM-DDTHH:MM:SSZ.
+  #
   # @returns [Course]
   def courses_api
     return unless authorized_action(@account, @current_user, :read_course_list)
+
+    starts_before = CanvasTime.try_parse(params[:starts_before])
+    ends_after = CanvasTime.try_parse(params[:ends_after])
 
     params[:state] ||= %w{created claimed available completed}
     params[:state] = %w{created claimed available completed deleted} if Array(params[:state]).include?('all')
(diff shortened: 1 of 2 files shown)


tim-schilling:
Is there anything else I need to do for this PR to move onto the next step?

tim-schilling:
@ccutrer Any chance you can give this PR a bump in Instructure's pipeline?

simonista:
Hi @tim-schilling, sorry to be slow getting back to you. Let me see if I can find someone to review this. I know we've had issues in the past with the performance of this endpoint, and so I'm a little worried about adding more filters that don't have an index associated with them. But let me see how the team feels and maybe we can do some performance testing on this.

In the past we've encouraged people who need change data to look at the live events stream, is that something that is possible in your use case?

tim-schilling:
We have looked at the live events stream in the past, but haven't felt like it provides the full functionality of what we need to start moving towards it. We're capturing the dates, term, name, sis_id for courses in particular. The major problems is that the live events stream (according to the docs) doesn't have an event for the dates or sis_id changing.

When we use the courses API endpoint, we end up iterating over almost the entire set of courses for schools since we can't use Canvas' completed state filter. This is because some schools would like our system to include courses before and/or after the courses active date range. That combined with the fact that the course dates can change, we are forced to iterate over the entire list each day to make sure we have a true list of active courses. The default params for this call on our end are:

```python
{
    "state[]": ["created", "claimed", "available"]
    "include[]": ["term", "total_students"],
    "blueprint": False,
}
```
We do try to use the `enrollment_term_id` param as often as possible, but it doesn't work for some schools.

Some of the worst cases is that we'll make over 200 requests paginating 10,000 courses to use 200 in our system. If we had the ability to specify a started after value for a course in this API call, we'd eliminate a lot of unnecessary requests.

I'd love to make this particular and other system interfaces more efficient for both of us, so if you need any more information please let me know.

simonista:
Feedback from the team that owns this controller is that you can move forward with this, but they had a suggested tweak to the SQL, which I will post inline.

app/controllers/accounts_controller.rb
```diff
@@ -609,6 +620,20 @@ def courses_api
       @courses = @courses.not_associated_courses
     end
 
+    if starts_before || ends_after
+      @courses = @courses.joins(:enrollment_term)
+      if starts_before
+        @courses = @courses.where("
+        (courses.start_at IS NULL AND enrollment_terms.start_at IS NULL)
+        OR courses.start_at <= ? OR enrollment_terms.start_at <= ?", starts_before, starts_before)
```
simonista: isn't it that if the course date is set, that takes precedence over the term date? if so, it should be

`"COALESCE(courses.start_at, enrollment_terms.start_at) <= ?"`

which you could then take to the next step:

`where("COALESCE(COALESCE(courses.start_at, enrollment_terms.start_at) <= ?, ?)", starts_before, true)`

(the result of a comparison between NULL and anything is NULL, so if they're both NULL, the outer COALESCE will do something - return TRUE. if either is NOT NULL, you get the result of the comparison)

tim-schilling: I don't think I can answer that question for you definitively. I was following the same business logic that's outlined here: https://github.com/instructure/canvas-lms/blob/a664cdb0b26bf9d4473c0204dba38fc73a34ece7/app/controllers/sis_api_controller.rb#L375-L384

However, looking at `Course`'s `:not_completed`, it's a logical `AND` that both the course `conclude_at` and term's `end_at` are less than `now`.

If I interpreted that wrong, I can change this.

ccutrer: while I wouldn't trust that controller to be definitive for anything, the not_complete scope on Course does behave as you suggest. it's still a bit weird to me, since enrollment dates (in EnrollmentDateBuilder#build prefers course dates over term dates (if the course is restricted, and the enrollment is not a teacher). But without the context of a specific user in a course, I don't know that we can do much better for saying when a course starts. So... this query is fine as is. Just please add to the documentation on these parameters how they behave to make it clear.

tim-schilling: If you feel it'd be better for the range to be inclusive of both the course and term's dates I'm fine changing it to that. I don't feel strongly in regards to the AND/OR nature of it or the idea of using `COALESCE`.

ccutrer: nah, it's fine as is.

tim-schilling: I've updated the documentation to explicitly mention that either the course's start_at or the term's start_at must be less than the value or both start_at values are `NULL`. And the same for `end_at`.

I also rebased on the latest version of master.

simonista:
Great, I'll get it pulled in and run through our full test suite.

tim-schilling:
Sounds good, thank you.

simonista:
6 months... better late than never, I suppose! :) Thanks again for the contribution.

PR #1487, opened 2019-08-08
Point Unsplash setup instructions to Wiki

lukechesser:
👋 I had messaged @claydiffrient to see if we could add some extra clarification for developers registering with the API.

Specifically for our API review team, It would be helpful to have users following this process note that they're creating a Canvas LMS application when applying for the Unsplash API. That will speed our team's review up.

I [updated the wiki entry](https://github.com/instructure/canvas-lms/wiki/Canvas-Integration#unsplash), so I thought it was best to point this line here.


SecurityInsanity:
Thank you for your contribution! :smile:

PR #1495, opened 2019-08-19
Updated Selenium Testing Doc

SamMalcolm:
```bash
brew install chromedriver
```
returns an error because there is no available formula present. Brew presents a cask `chromedriver` as an alternative. So I've updated the bash command in the docs to reflect


claydiffrient:
This was merged as 3f29fb0f2c94c75a120a96b0c0c487393f9014a2.

PR #1501, opened 2019-08-27
Ignore spacers when activating submit (Quizzes)

leo60228:
Previously, the Submit Quiz button would never be activated (blue) if the quiz had any spacers. This changes `li:not(.answered)` to `li:not(.answered, .text_only)` in order to solve this.

Test Plan:
  - Make a quiz with one question and one spacer. Ensure that the Submit Quiz button is activated when the question has an answer.


simonista:
Hey @leo60228, thanks for the pull request. I checked it out and it looks good to me. Could you try cla-assistant.io one more time? If not I can email you a copy of the CLA to sign another way. Thanks!

PR #1510, opened 2019-10-03
fix russian typo

DoubleJarvis:
There is a typo in Russian translation, this fixes it


ccutrer:
Canvas' localization workflow involves third party components, and are automatically imported, so we can't commit fixes directly to Canvas or they'll be overwritten. This particular fix has been imported into the upstream system, and should be reflected in Canvas within the next day or so.

PR #1527, opened 2019-11-05
Make make_unique_filename deal with extensions starting with digits

christopher-b:
Fix an issue where Attachment#make_unique_filename would improperly rename files with extensions starting with digits (e.g. Rhino .3dm files). A file `myfile.3dm` would be renamed to `myfile.3dm-1` rather than `myfile-1.3dm`.

 Test Plan:
   - Upload a file called `myfile.3dm`
   - Upload a different file with the same name
   - Ensure the new file is named `myfile-1.3dm`

### DIFF
--- a/app/models/attachment.rb
+++ b/app/models/attachment.rb
@@ -1813,7 +1813,7 @@ class Attachment < ActiveRecord::Base
     addition = 1
     dir = File.dirname(filename)
     dir = dir == "." ? "" : "#{dir}/"
-    extname = filename[/(\.[A-Za-z][A-Za-z0-9]*)+$/] || ''
+    extname = filename[/(\.[A-Za-z][A-Za-z0-9]*)*(\.[A-Za-z0-9]*)$/] || ''
     basename = File.basename(filename, extname)
 
     until block.call(new_name = "#{dir}#{basename}-#{addition}#{extname}")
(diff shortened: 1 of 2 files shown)


simonista:
Hey @christopher-b, thanks for the PR. Looking at the history here, it looks like originally we were using File.extname, and then switched to the regex form to better handle things like `.tar.gz` or `.html.erb`.

Is there a specific reason you are only allowing starting digits in the last segment of the extension? For example, something like `blah.3dm.example` would become `blah.3dm-1.example`. I'm not sure there's really any standard to follow here...

One option would just be to split on the first `.` (but probably want to special case filenames starting with `.`). Or another option would be to keep the regex as-is but allow starting digits in all segments.

Thoughts?

simonista:
Aha! Found more historical context from the ticket:

> what I wanted to avoid was, if someone named a file "This is my submission. I have no idea what I'm doing.", I don't want to treat the entire second sentence as an extension. I could say neither dot nor whitespace, I suppose.

> but we have other problems here that lead me to wonder if this is worth doing at all.  say someone names a file "Section 11.5.doc".  the renamed file should be "Section 11.5-1.doc", and not "Section 11-1.5.doc".

I'm not sure how to handle both `blah.3dm` and `Section 11-1.5.doc` well...

christopher-b:
Thanks for taking a look, @simonista. You make a good point. And yes, I was trying to account for cases like `section 11.5.doc`, as described in the specs. 

My approach was a compromise that would satisfy both cases (`blah.3dm` and `Section 11.5.doc`) but not the one you're describing (`blah.3dm.example `). I think it's an incremental improvement but not perfect. This change does make the specs pass, but wonder what the real-world impact would be.

simonista:
Oh ha, I should have just gone and looked at the tests! Sounds good, I can see this being a good compromise and incremental step forward. Let me run it by the team in charge of attachments and see what they think as well.

PR #1532, opened 2019-11-19
Fix adding students to a conference

bfcoder:
Fix backend to refresh the conference data to get an updated list of
users.

Test Plan:
  - Create a conference and only add one student.
  - Now edit that conference and add another student.
  - If you edit that conference again, that second student should have a checkmark indicating they are an invitee to the conference now.

### DIFF
--- a/app/controllers/conferences_controller.rb
+++ b/app/controllers/conferences_controller.rb
@@ -256,6 +256,7 @@ class ConferencesController < ApplicationController
             @conference.add_invitee(u)
           end
           @conference.save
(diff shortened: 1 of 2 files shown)


simonista:
Hey @bfcoder, thanks for the pull request. I did some digging and ended up fixing this in a slightly different way (I'll link here once it's merged), but I appreciate you writing the failing test that made debugging easier!

bfcoder:
Cool, yeah I'm interesting to see the solution you came up with. And you're welcome for the test!

PR #1538, opened 2019-11-27
Fix typo in student_planner_spec (Selenium)

LukeSerne:
There was a typo in `canvas-lms/spec/selenium/dashboard/planner/student_planner_spec.rb`. That has been fixed now. This should not change any behaviour and thus, we can be sure that this does not introduce any new bugs.

Test Plan:
  - No special testing needed, since no behaviour was changed.


simonista:
Thanks for the PR!

PR #1542, opened 2019-12-03
Fix documented return types of some Assignment fields

### LINKED ISSUE #1539: Bug in (or documentation error for) Assignments REST API.
### Summary:
 - `Assignment.integration_data` returns `"object"` instead of (as documented) a `"string"`: https://github.com/instructure/canvas-lms/blob/eaa9fd10071711f9db248ed08c0bee33297fa3e6/app/controllers/assignments_api_controller.rb#L417-L421
 - `Assignment.possible_points` returns `"number"` instead of (as documented an `"integer"`: https://github.com/instructure/canvas-lms/blob/eaa9fd10071711f9db248ed08c0bee33297fa3e6/app/controllers/assignments_api_controller.rb#L426-L430

### Steps to reproduce:

```
curl -H "Authorization: Bearer ${CANVAS_TOKEN}" ${CANVAS_URL}/api/v1/courses/:course_id/assignments/:assignment_id
```
returns
```
{...,"points_possible":3.0,...,"integration_data":{},...}
```

### Additional notes:

If this is just an error in the docs I can submit a PR, but would be good if someone could confirm whether this is a mistake in the docs or the implementation.

fredrikekre:
Fix documented return types of `Assignment.points_possible` and `Assignment.integration_data`, fixes  #1539.

### DIFF
--- a/app/controllers/assignments_api_controller.rb
+++ b/app/controllers/assignments_api_controller.rb
@@ -415,8 +415,8 @@
 #           "description" : "(optional, Third Party unique identifier for Assignment)"
 #         },
 #         "integration_data": {
-#           "example": "12341234",
-#           "type" : "string",
+#           "example": {"5678": "0954"},
+#           "type" : "object",
 #           "description" : "(optional, Third Party integration data for assignment)"
 #         },
 #         "muted": {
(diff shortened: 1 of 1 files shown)


app/controllers/assignments_api_controller.rb
```diff
@@ -415,8 +415,8 @@
 #           "description" : "(optional, Third Party unique identifier for Assignment)"
 #         },
 #         "integration_data": {
-#           "example": "12341234",
-#           "type" : "string",
+#           "example": {"5678": "0954"},
```
fredrikekre: I don't know if this is a "good" example, I took it from some other example for `integration_data` in another object example.

simonista: I'm not super familiar either, but in looking at the code it looks like keys and values can be anything, whatever is meaningful to the third party that is using it, I guess.

simonista:
Merged in 8e96b8b

PR #1560, opened 2020-01-13
Make add app dialog remain open on document click

claydiffrient:
This dialog is large enough and complicated enough that it stinks when you're inputting
lots of data and happen to brush your trackpad or something causing it to close and lose
data. :(

Test Plan:
  - Go to external apps and add an app
  - When the modal to add the app opens, click outside of it.
  - The modal should stay open.


claydiffrient:
@westonkd Someone from your team should take a look at this probably :)

simonista:
+1 from eng/product. running through jenkins

PR #1561, opened 2020-01-14
Prevent page error when Dashcard position contains string

jojo787898:
When set_dashboard_positions is given a string instead of an integer, they are happily stored in the user record.
Later when the dashboard is rendered, Ruby tries to sort the dashcards by :position and this means that a string is compared against an int, crashing the application with "comparison of String with 2 failed" or something similar.

Test Plan:
- Added new test case in dashboard_helper_spec.rb as suggested by @claydiffrient in the previous pull request #1407 

```
Jonathans-MacBook-Pro:canvas-lms jonathannovecio$ docker-compose run --rm web bundle exec rspec spec/helpers/dashboard_helper_spec.rb
Starting canvas-lms_postgres_1 ... done
Starting canvas-lms_redis_1    ... done
/home/docker/.gem/ruby/2.4.0/gems/soap4r-ruby1.9-2.0.5/lib/soap/mapping/encodedregistry.rb:150: warning: constant ::Fixnum is deprecated
/home/docker/.gem/ruby/2.4.0/gems/soap4r-ruby1.9-2.0.5/lib/soap/mapping/encodedregistry.rb:216: warning: constant ::Fixnum is deprecated
DEPRECATION WARNING: `secrets.secret_token` is deprecated in favor of `secret_key_base` and will be removed in Rails 6.0. (called from <top (required)> at /usr/src/app/config/environment.rb:28)
Run options: exclude {:pact_live_events=>true}

Randomized with seed 47507
........

Finished in 9.01 seconds (files took 27.31 seconds to load)
8 examples, 0 failures

Randomized with seed 47507

```

### DIFF
--- a/app/presenters/course_for_menu_presenter.rb
+++ b/app/presenters/course_for_menu_presenter.rb
@@ -42,7 +42,7 @@ class CourseForMenuPresenter
       id: course.id,
       isFavorited: course.favorite_for_user?(@user) && @user.account.feature_enabled?(:unfavorite_course_from_dashboard),
       image: course.feature_enabled?(:course_card_images) ? course.image : nil,
-      position: @user.dashboard_positions[course.asset_string] || nil,
+      position: @user.dashboard_positions[course.asset_string].to_i || nil,
     }.tap do |hash|
       if @opts[:tabs]
         tabs = course.tabs_available(@user, {
--- a/app/presenters/course_for_menu_presenter.rb
+++ b/app/presenters/course_for_menu_presenter.rb
@@ -42,7 +42,12 @@ class CourseForMenuPresenter
       id: course.id,
       isFavorited: course.favorite_for_user?(@user) && @user.account.feature_enabled?(:unfavorite_course_from_dashboard),
       image: course.feature_enabled?(:course_card_images) ? course.image : nil,
-      position: @user.dashboard_positions[course.asset_string].to_i || nil,
+      position: if @user.dashboard_positions[course.asset_string].present? == true
+                    @user.dashboard_positions[course.asset_string].to_i
+                else
+                    nil
+                end
+       # @user.dashboard_positions[course.asset_string].to_i || nil,
     }.tap do |hash|
       if @opts[:tabs]
         tabs = course.tabs_available(@user, {
--- a/spec/helpers/dashboard_helper_spec.rb
+++ b/spec/helpers/dashboard_helper_spec.rb
@@ -97,7 +97,7 @@ describe DashboardHelper do
         mapped_courses = map_courses_for_menu(courses)
         expect(mapped_courses.map {|h| h[:id]}).to eq [course3.id, course2.id, course1.id]
       end
-      
+
       it "handles sorting even when positions are strings" do
         course1 = @account.courses.create!
         course2 = @account.courses.create!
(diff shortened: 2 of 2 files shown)


app/presenters/course_for_menu_presenter.rb
```diff
@@ -42,7 +42,7 @@ def to_h
       id: course.id,
       isFavorited: course.favorite_for_user?(@user) && @user.account.feature_enabled?(:unfavorite_course_from_dashboard),
       image: course.feature_enabled?(:course_card_images) ? course.image : nil,
-      position: @user.dashboard_positions[course.asset_string] || nil,
+      position: @user.dashboard_positions[course.asset_string].to_i || nil,
```
simonista: I don't think this is quite what we want, because `nil.to_i == 0` in ruby, so if the user doesn't have a position set for a course, this will make it 0, instead of falling back to the `|| nil` case.

I think you'll have to do something like:
```
if @user.dashboard_positions[course.asset_string].present?
  @user.dashboard_positions[course.asset_string].to_i
else
  nil
end
```
or 
```
pos = @user.dashboard_positions[course.asset_string]
pos.present? ? pos.to_i : nil
```

jojo787898: Thanks for the suggestion. I added the code that you suggested and updated the test.

Jonathans-MacBook-Pro:canvas-lms jonathannovecio$ docker-compose run --rm web bundle exec rspec spec/helpers/dashboard_helper_spec.rb
Starting canvas-lms_redis_1 ... done
Starting canvas-lms_postgres_1 ... done
/home/docker/.gem/ruby/2.4.0/gems/soap4r-ruby1.9-2.0.5/lib/soap/mapping/encodedregistry.rb:150: warning: constant ::Fixnum is deprecated
/home/docker/.gem/ruby/2.4.0/gems/soap4r-ruby1.9-2.0.5/lib/soap/mapping/encodedregistry.rb:216: warning: constant ::Fixnum is deprecated
DEPRECATION WARNING: `secrets.secret_token` is deprecated in favor of `secret_key_base` and will be removed in Rails 6.0. (called from <top (required)> at /usr/src/app/config/environment.rb:28)
Run options: exclude {:pact_live_events=>true}

Randomized with seed 28838
.........

Finished in 9.42 seconds (files took 27.48 seconds to load)
9 examples, 0 failures

Randomized with seed 28838

jojo787898:
Hi, did my last commit address your concern @simonista ?

app/presenters/course_for_menu_presenter.rb
```diff
@@ -42,7 +42,12 @@ def to_h
       id: course.id,
       isFavorited: course.favorite_for_user?(@user) && @user.account.feature_enabled?(:unfavorite_course_from_dashboard),
       image: course.feature_enabled?(:course_card_images) ? course.image : nil,
-      position: @user.dashboard_positions[course.asset_string] || nil,
+      position: if @user.dashboard_positions[course.asset_string].present? == true
```
simonista: A few small nitpicks:
- `== true` is redundant and unnecessary here because `.present?` already returns true or false
- I don't like seeing a full if/else/end block nested inside a hash definition like this. Just personal preference. I'd either extract this out into a variable above the hash definition, or extract `@user.dashboard_positions[course.asset_string]` into a shorter variable and then use an inline ternary expression (like on the previous line).

Also, the commented out line of code on 50 should just be removed.

These are all small minor things, and I have to import this into our internal system to run tests anyway, so I'll just fix them up when I do that. Overall the logic looks sound to me. We'll see what the tests think :)

PR #1572, opened 2020-02-09
[.rubocop.yml] Upgraded namespaces for compatability with latest rubocop

SamuelMarks:
It might also be a good idea to sort this YAML file, but I didn't want to do this without your approval.

Working with rubocop version 0.79.0.

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -8,7 +8,7 @@ AllCops:
   TargetRubyVersion: 2.3
 
 # our style changes: disabling style rules we aren't interested in
-Layout/AlignParameters:
+Layout/ParameterAlignment:
   Enabled: false
 Layout/ElseAlignment:
   Enabled: false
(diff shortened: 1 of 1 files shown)


simonista:
Hi @SamuelMarks , sorry to be so slow responding to this.

We currently have rubocop pinned at 0.52.1, will these changes still work with that version? (see https://github.com/instructure/canvas-lms/blob/master/Gemfile.d/test.rb#L41)

SamuelMarks:
Give it a shot?

simonista:
I think if that's the attitude, we'll probably turn this down for now. We're still evaluating if/how we want to move forward with Rubocop, and so in order to accept this I'd need to know for sure that it works with our current version.

SamuelMarks:
@simonista I've merged in master, and carefully ensured that only the changes that are still valid were added to the `.rubocop.yml` file.

PR #1585, opened 2020-03-03
Update documentation for CalenderEventsApiController#set_course_timeable_events

MWndrs:
events[][title] is a allowed parameter for the `POST CalenderEventsApiController#set_course_timetable_events` call, however the API documentation does not list this. Since we want to use this parameter we would like the API documentation to be up-to-date in this regard.

Test Plan:
  - Check /doc/api/calendar_events.html#method.calendar_events_api.set_course_timetable_events to see if events[][title] is listed as parameter
  - Check you can actually set the events title using this call :>. (relevant code at app/models/courses/timetable_event_builder.rb:162)


simonista:
Thanks for the PR, and sorry for the slow response. I'll get this pulled in.

PR #1598, opened 2020-04-02
Courses can also be looked up by the integrationID

buckett:
You can lookup courses by an integration ID with a URL like this:

https://instance.instructure.com/api/v1/courses/sis_integration_id:test-int-123


buckett:
@slaughter550 Thanks for looking into this.

slaughter550:
@buckett I had to rebase this commit. CI is running again. Will let you know shortly.

slaughter550:
@buckett your change has been merged in! Thanks so much for being a contributor!

PR #1616, opened 2020-04-29
Add ERB-rendering of `config/logging.yml`

ktham:
## Changes
* Use built-in Rails helper `config_for` for automatic ERB-rendering and YAML-parsing of `config/logging.yml`

This keeps it consistent with how Rails parses other config files in that directory and also makes it easier to specify parameters using environment variables.

See [railties/lib/rails/application.rb#L207](https://github.com/rails/rails/blob/d9bfafd9b8af6d3ee099d6a3d431744b70666aac/railties/lib/rails/application.rb#L207) for more information on `config_for`

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -65,8 +65,8 @@ module CanvasRails
     # Run "rake -D time" for a list of tasks for finding time zone names. Comment line to use default local time.
     config.time_zone = 'UTC'
 
-    log_config = File.exist?(Rails.root+"config/logging.yml") && YAML.load_file(Rails.root+"config/logging.yml")[Rails.env]
-    log_config = { 'logger' => 'rails', 'log_level' => 'debug' }.merge(log_config || {})
+    log_config = Rails.application.config_for(:logging) rescue {}
+    log_config = { 'logger' => 'rails', 'log_level' => 'debug' }.merge(log_config)
     opts = {}
     require 'canvas_logger'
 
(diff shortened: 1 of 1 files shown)


ktham:
@djbender What do you think? Any questions?

The reason for this PR is that I'd like to be able to set log levels using an environment variable. This PR also has the added benefit of re-using a built-in Rails function for loading and parsing yaml configs vs implementing your own config file lookup logic

config/application.rb
```diff
@@ -65,8 +65,8 @@ class Application < Rails::Application
     # Run "rake -D time" for a list of tasks for finding time zone names. Comment line to use default local time.
     config.time_zone = 'UTC'
 
-    log_config = File.exist?(Rails.root+"config/logging.yml") && YAML.load_file(Rails.root+"config/logging.yml")[Rails.env]
-    log_config = { 'logger' => 'rails', 'log_level' => 'debug' }.merge(log_config || {})
+    log_config = Rails.application.config_for(:logging) rescue {}
```
slaughter550, line 68: @ktham we would still want to see if the file exists and not use rescue to gobble parsing errors, etc. 

`File.exist?(Rails.root+"config/logging.yml") && Rails.application.config_for(:logging)` is probably best and leave the line underneath as is.

Would that still work for your use case?

slaughter550 (changes requested):

config/application.rb
ktham: Sure, done

slaughter550 (approved):

slaughter550:
@ktham Thanks for your contribution! It has been merged in https://github.com/instructure/canvas-lms/commit/d94ba37a70e44efb411aeb1531af5a31d969fe35

PR #1661, opened 2020-06-30
Change canvas_quizzes dependencies to use https instead of git protocol

grahamb:
The canvas_quizzes dependency for grunt-template-jasmine-requirejs usethe git protocol (`git://`). This is an issue when trying to build in a Docker environment AND when behind a corporate firewall. Git can be configured to prefer `https://` instead of `git://`; however, this configuration takes place in the host's gitconfig file and isn't visible to the git
implementation inside a container (unless you mount your host's gitconfig inside the container, which may or may not be desirable).

This commit changes canvas_quizzes' package.json file to use `https://` instead of `git://`. This resolves the issue of not being able to fetch the dependency from inside a container whilst behind a firewall since Docker will pick up the `https_proxy` environment variables from the host and use the proxy to fetch the dependency via https.

We have been using this change at SFU for many weeks without issue.

**Test plan:**

- check out the repo
- run `yarn`

Expected result:

- all dependencies should be fetched without **issue**

- check out the repo
- run `yarn`

**Expected result:**

- all dependencies should be fetched without issue

### DIFF
--- a/client_apps/canvas_quizzes/package.json
+++ b/client_apps/canvas_quizzes/package.json
@@ -54,7 +54,7 @@
     "grunt-contrib-watch": "1.0.0",
     "grunt-jsduck": "1.0.1",
     "grunt-sass": "^2.1.0",
-    "grunt-template-jasmine-requirejs": "git://github.com/amireh/grunt-template-jasmine-requirejs.git#0.2.0",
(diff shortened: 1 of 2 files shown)


maths22:
Merged in 79b29ddfd2e4c906fd8602b2d51c7d0235ce7275

PR #1669, opened 2020-07-11
Add score statistics to the assignments API as an 'include' option

ndrewh:
Provides an API for the information in the grade distribution graphs shown in the web UI on the 'Grades' screen to students.

The permissions model is identical: students can view grade grade distribution (min, max, mean) only if all of the following:
- The 'hide_distribution_graphs' setting is NOT enabled by the instructor in the course's settings
- There are at least 5 graded submissions
- The student's submission is graded, and the grade is visible to them.

Instructors (or anyone with the read_admin permission) can always view distributions when they are looking at a student with a graded assignment.

Test Plan:
  - Make API requests to the GET endpoints for individual assignments, assignment groups, and for listing all assignments for a course
  - Test these API endpoints with and without the include[]=score_statistics flag.
  - Verify that the statistics match those shown in the grade distribution graphs on in the web UI
  - Verify that the statistics are only available in the API when there are 5 graded submissions and when not disabled by the instructor (this behavior should match the distribution graphs in the web UI)

My ultimate goal is to add this functionality to the iOS app, once this gets merged.

### DIFF
--- a/app/controllers/assignments_api_controller.rb
+++ b/app/controllers/assignments_api_controller.rb
@@ -241,6 +241,29 @@
 #       }
 #     }
 #
+# @model ScoreStatistic
+#     {
+#       "id": "ScoreStatistic",
+#       "description": "Used by Assignment model",
+#       "properties": {
+#         "min": {
+#           "description": "Min score",
+#           "example": 1,
+#           "type": "integer"
+#         },
+#         "max": {
+#           "description": "Max score",
+#           "example": 10,
+#           "type": "integer"
+#         },
+#         "mean": {
+#           "description": "Mean score",
+#           "example": 6,
+#           "type": "integer"
+#         }
+#       }
+#     }
+#
 # @model Assignment
 #     {
 #       "id": "Assignment",
@@ -605,6 +628,10 @@
 #           "description": "Whether the assignment has manual posting enabled. Only relevant for courses using New Gradebook.",
 #           "example": true,
 #           "type": "boolean"
+#         },
+#         "score_statistics": {
+#           "description": "(Optional) If 'score_statistics' and 'submission' are included in the 'include' parameter and statistics are available, includes the min, max, and mode for this assignment",
+#           "$ref": "ScoreStatistic"
 #         }
 #       }
 #     }
@@ -619,7 +646,7 @@ class AssignmentsApiController < ApplicationController
 
   # @API List assignments
   # Returns the paginated list of assignments for the current course or assignment group.
-  # @argument include[] [String, "submission"|"assignment_visibility"|"all_dates"|"overrides"|"observed_users"|"can_edit"]
+  # @argument include[] [String, "submission"|"assignment_visibility"|"all_dates"|"overrides"|"observed_users"|"can_edit"|"score_statistics"]
   #   Optional information to include with each assignment:
   #   submission:: The current user's current +Submission+
   #   assignment_visibility:: An array of ids of students who can see the assignment
@@ -627,6 +654,7 @@ class AssignmentsApiController < ApplicationController
   #   overrides:: An array of +AssignmentOverride+ structures
   #   observed_users:: An array of submissions for observed users
   #   can_edit:: an extra Boolean value will be included with each +Assignment+ (and +AssignmentDate+ if +all_dates+ is supplied) to indicate whether the caller can edit the assignment or date. Moderated grading and closed grading periods may restrict a user's ability to edit an assignment.
+  #   score_statistics:: An object containing min, max, and mean score on this assignment. This will not be included for students if there are less than 5 graded assignments or if disabled by the instructor. Only valid if 'submission' is also included.
   # @argument search_term [String]
   #   The partial title of the assignments to match and return.
   # @argument override_assignment_dates [Boolean]
@@ -806,6 +834,10 @@ class AssignmentsApiController < ApplicationController
 
       preloaded_attachments = api_bulk_load_user_content_attachments(assignments.map(&:description), @context)
 
+      if include_params.include?('score_statistics')
+        ActiveRecord::Associations::Preloader.new.preload(assignments, :score_statistic)
+      end
+
       hashes = []
       hashes = assignments.map do |assignment|
 
@@ -824,7 +856,8 @@ class AssignmentsApiController < ApplicationController
                         bucket: params[:bucket],
                         include_overrides: include_override_objects,
                         preloaded_user_content_attachments: preloaded_attachments,
-                        include_can_edit: include_params.include?('can_edit')
+                        include_can_edit: include_params.include?('can_edit'),
+                        include_score_statistics: include_params.include?('score_statistics')
                         )
       end
       hashes
@@ -833,10 +866,11 @@ class AssignmentsApiController < ApplicationController
 
   # @API Get a single assignment
   # Returns the assignment with the given id.
-  # @argument include[] [String, "submission"|"assignment_visibility"|"overrides"|"observed_users"|"can_edit"]
+  # @argument include[] [String, "submission"|"assignment_visibility"|"overrides"|"observed_users"|"can_edit"|"score_statistics"]
   #   Associations to include with the assignment. The "assignment_visibility" option
   #   requires that the Differentiated Assignments course feature be turned on. If
   #   "observed_users" is passed, submissions for observed users will also be included.
+  #   For "score_statistics" to be included, the "submission" option must also be set.
   # @argument override_assignment_dates [Boolean]
   #   Apply assignment overrides to the assignment, defaults to true.
   # @argument needs_grading_count_by_section [Boolean]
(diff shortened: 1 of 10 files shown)


ndrewh:
Android: https://github.com/instructure/canvas-android/pull/888
iOS: https://github.com/instructure/canvas-ios/pull/1421

ndrewh:
Is there anything I need to do on this PR? There is a PR for each of the mobile apps that depend on this change.

Thanks!

armstrongnate:
@ndrewh We have created a ticket internally to review this PR. Thank you for your contribution and for your patience!

ndrewh:
Hi @spencerolson 

I was able to fix essentially all the tests with a 1 line fix. (Feel free to squash into previous commit, but I figured you'd be rebasing before this hits master anyways).

My apologies for not running all of the tests prior to submitting the PR in the first place. I don't have the computational power to run *all* the tests, so I only ran the test suites that seemed related. I have now fixed and re-run all of the test cases you listed, and **they all now pass with the following exceptions**:

Failing tests I couldn't reproduce on my machine
----

```
spec.selenium.assignments.assignments_direct_share_import_spec.assignments with direct share FF ON copy to copy tray lists course modules
```

Additional tests in that suite I had fail, some of which you did not (i'm pretty sure these are all unrelated to this change)
-----
```
rspec ./spec/selenium/assignments/assignments_direct_share_import_spec.rb:161 # assignments with direct share FF ON send to can send an item to another instructor
rspec ./spec/selenium/assignments/assignments_direct_share_import_spec.rb:172 # assignments with direct share FF ON send to received item can preview received item
rspec ./spec/selenium/assignments/assignments_direct_share_import_spec.rb:178 # assignments with direct share FF ON send to received item can be imported into a course
rspec ./spec/selenium/assignments/assignments_direct_share_import_spec.rb:140 # assignments with direct share FF ON copy to copied assignment is present in destination course
```

Tests that fail because of a local issue I don't know how to fix
------

I can't verify that the 1 dashboard selenium failure is fixed because I have an unrelated error. I tried changing the theme back to default. they still fail with the same error. I don't think this error is in any way relevant to this PR. 
```
rspec ./spec/selenium/dashboard_spec.rb: (literally all of the tests in this suite)

      Failure/Error: raise RuntimeError, javascript_errors.map(&:message).join("\n\n")
      
      RuntimeError:
        webpack:///./packages/js-utils/lib/filterUselessConsoleMessages.js? 51:22 "Warning: [themeable] A theme registry has already been initialized. Ensure that you are importing only one copy of '@instructure/ui-themeable'." ""
      Shared Example Group: "load events list" called from ./spec/selenium/dashboard_spec.rb:373
      # ./spec/selenium/common.rb:282:in `block (2 levels) in <top (required)>'
      # ./spec/support/great_expectations.rb:63:in `run_after_example'
      # ./spec/support/spec_time_limit.rb:46:in `block in enforce'
      # ./spec/support/spec_time_limit.rb:45:in `enforce'
      # ./spec/spec_helper.rb:362:in `block (2 levels) in <top (required)>'
      # ./spec/support/blank_slate_protection.rb:48:in `block in run'
      # ./spec/support/blank_slate_protection.rb:75:in `disable'
      # ./spec/support/blank_slate_protection.rb:48:in `run'

```

I tried rebasing on master and rebuilding my entire docker-compose setup, to no avail in fixing the above tests. 


I was able to verify that all of the other tests (including the ~40 from  `spec.models.user_learning_object_scopes_spec`, `spec.apis.v1.todo_items_api_spec`, and `spec.presenters.to_do_list_presenter_spec`) are fixed. I suspect the same root cause was failing the other tests on your machine, but for the reasons above I cannot verify this on my own machine.

Let me know if there are any other issues.

Thanks,
@ndrewh

spencerolson:
Hi @ndrewh , thanks for the thorough response! I'll go ahead and run this through our automated test suite again, and then report back here.

> Feel free to squash into previous commit or rebase, but I figured you'd be rebasing before this hits master anyways so I didn't want to disrupt your workflow.

Thanks, I'll go ahead and squash this into the previous commit. And you're right, I'll rebase this before it hits master.

> Tests that failed for you, but I couldn't reproduce on my machine

> Additional tests in that suite I had fail, some of which you did not (i'm pretty sure these are all unrelated to this change)

I'll keep an eye on these; it's possible they are flaky tests and that the failures are unrelated to this change.

> I can't verify that the 1 dashboard selenium failure you had is fixed because I have an unrelated error. I tried changing the theme back to default. they still fail with the same error. I don't think this error is in any way relevant to this PR.

Agree, this seems like an issue with the local setup. I did some digging around to see if any others had run into this issue to see if I could offer you a suggestion for a fix, but not seeing anything, sorry.

Thanks,
@spencerolson

spencerolson:
@ndrewh great news, it passed the test suite! I'll go ahead and get this scheduled for code review and post back here with feedback, if any.

lib/api/v1/assignment.rb
```diff
@@ -356,8 +356,18 @@ def assignment_json(assignment, user, session, opts = {})
       if submission.is_a?(Array)
         ActiveRecord::Associations::Preloader.new.preload(submission, :quiz_submission) if assignment.quiz?
         hash['submission'] = submission.map { |s| submission_json(s, assignment, user, session, assignment.context, params) }
+        should_show_statistics = opts[:include_score_statistics] && submission.any? { |s| can_show_statistics?(assignment, s, user) }
```
spencerolson, line 359: It looks like this will lead to an N+1 query problem because Submission#hide_grade_from_student? (invoked on line 406) will make queries. The lines in that method that are currently executing queries are:

1. `return muted_assignment? unless assignment.course.post_policies_enabled?`

Suggestion: the "Post Policies" feature is enabled for everyone in Canvas now, so this line can be safely removed in this commit (we plan to follow up with a commit to remove all checks for this feature flag).

2. `if assignment.post_manually?`

Since the `assignment` association isn't loaded on the submission, this will result in a query to fetch the assignment. However, conveniently, we have the assignment available here, so we just need to manually "load" it onto the submission. On line 405, we should be able to add:

```
submission.assignment = assignment
```

. This way, the `if assignment.post_manually?` line will no longer execute a query to get the assignment (there's some work we'll probably want to do on our side to do make the post_manually? method more performant, but that's outside the scope of your commit).

lib/api/v1/assignment.rb
```diff
@@ -391,6 +401,19 @@ def assignment_json(assignment, user, session, opts = {})
...
+    # To show statistics, the student's grade must be visible to the student
+    return false unless (submission.score.present? && !submission.hide_grade_from_student?)
+
+    # and the assignment must have points_possible > 0,
+    return false unless (assignment.points_possible.present? && assignment.points_possible > 0)
+
+    # Students can only see statistics when count >= 5 and not disabled by the instructor
+    # Instructor can see statistics at any time.
+    count = assignment.score_statistic&.count || 0
+    assignment.context.grants_right?(user, :read_as_admin) || (count >= 5 && !assignment.context.hide_distribution_graphs)
```
spencerolson, line 414: for the invocation of this method on line 359, where we invoke this once-per-submission, it looks like these lines are going to be doing duplicated work, since it's going to be the same assignment each time this method is run.

we suggest breaking this method out into 2 separate methods: one for the submission-specific-work, and one for the assignment-specific-work. that way, the invocation on line 359 can call the submission-specific-work method once-per-submission, but only call the assignment-specific-work method one single time.

spencerolson, line 414: It looks like the "count >= 5" logic here is based on the logic in `GradeSummaryAssignmentPresenter#hide_distribution_graphs?` , is that right? It'd be great if we could find a way to abstract the logic into a shared place, so that we can DRY things up a bit (also, that way, if we decide down the road that we want to change the threshold from 5 to something else, we have one place to change it).

ndrewh:
I made the requested changes, squashed, and rebased. I only had the time to run a few of the test suites.

Thanks,
@ndrewh

lib/api/v1/assignment.rb
ndrewh: Yeah that's where I got the logic from. I moved the logic to the assignment and submission model and it's shared now.

app/controllers/assignments_api_controller.rb
```diff
@@ -806,6 +834,10 @@ def get_assignments(user)
 
       preloaded_attachments = api_bulk_load_user_content_attachments(assignments.map(&:description), @context)
 
+      if include_params.include?('statistics')
```
spencerolson, line 837: Looks like this should be 'score_statistics' instead of 'statistics', right?

ndrewh: Yeahhhh that's supposed to be score_statistics. Feel free to make that change - no need to wait for me. I don't know if I somehow messed that up in the most recent push or if it was like that all along, but good catch.

ndrewh:
@spencerolson I fixed that line. I also added tests for observer users, as we weren't testing that case at all before.

spencerolson:
@ndrewh Oh nice, thanks for adding additional tests for observers! Will run this through our test suite + code review again.

spencerolson:
Hi @ndrewh , your commit has been merged to the `master` branch and can be found here: https://github.com/instructure/canvas-lms/commit/92a038cd53a58dd972ba85576bd4f680133470d4 . Thanks for your contribution, and for helping to make Canvas better. We really appreciate it!

PR #1700, opened 2020-09-14
Remove cache-with-validation from set_no_cache_headers to prevent Safari caching images

karendolan:
This update is to assist the Safari CORS issues with Canvas uploaded images that are served from an AWS S3 bucket. Safari errors out with CORS error. This is only when Safari makes a CORS OPTIONS validate request on the cached image, that other browsers do not make.

> FetchEvent.respondWith received an error: TypeError: Cross-origin redirection to https://instructure-uploads.s3.amazonaws.com/[the bucket link and params] denied by Cross-Origin Resource Sharing policy: Origin [the Canvas LMS DNS] is not allowed by Access-Control-Allow-Origin.

Only Safari is caching and making the validate request. Chrome and Firefox do not.
According to the Mozilla Cache-Control post, the "no-cache,no-store" is contradictory:

> **GOOD:** "Cache-Control: no-store"
> **BAD**: "Cache-Control: no-cache,no-store"
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Cache-Control

> no-cache:
> The response may be stored by any cache, even if the response is normally non-cacheable. However, the stored response MUST always go through validation with the origin server first before using it, therefore, you cannot use no-cache in-conjunction with immutable. If you mean to not store the response in any cache, use no-store instead. This directive is not effective in preventing caches from storing your response.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1283,7 +1283,7 @@ class ApplicationController < ActionController::Base
 
   def set_no_cache_headers
     response.headers["Pragma"] = "no-cache"
-    response.headers["Cache-Control"] = "no-cache, no-store"
+    response.headers["Cache-Control"] = "no-store"
   end
 
   def set_page_view


karendolan:
This solution is not optimal, but it might alleviate the cross-site image cookie cache problem for Safari 13.1 users. I believe the old "no-cache" addition to the "no-store" was used to support older IE browsers.

karendolan:
"All browsers currently will not reuse a response with no-cache" implying that FireFox and Chrome will follow Safari on this. https://github.com/whatwg/fetch/issues/1088 

 "For images, the memory cache is checked before the service worker, and if there's a memory cache hit, a fetch event is never fired. " Only Safari appears to be doing this for the Canvas redirect images. https://github.com/GoogleChrome/workbox/issues/2540

Alternatively, adding a terminating solidus (forward slash) to the Canvas image src redirect stops Safari from treating the image src as a normal cached resource. Safari successfully retrieves the image on each refresh when the image src redirect URL contains the terminating solidus.

karendolan:
For the server side:

The following curl checks Access-Control-Allow-Methods for instructure-uploads.s3.amazonaws.com as recommended in AWS S3 ticket referenced below. The OPTIONS method is not listed as one of the accepted methods. It's possible that the Safari preflight OPTIONS method is not allowed on https://instructure-uploads.s3.amazonaws.com and is affecting the preflight request on cached CORS images.

Access-Control-Allow-Methods: GET, PUT, POST, OPTIONS  <---- missing ?

```
$ curl -H "origin: any" -v "https://instructure-uploads.s3.amazonaws.com"

> GET / HTTP/1.1
> Host: instructure-uploads.s3.amazonaws.com

< Access-Control-Allow-Origin: *
< Access-Control-Allow-Methods: GET, PUT, POST
< Access-Control-Max-Age: 3000
< Vary: Origin, Access-Control-Request-Headers, Access-Control-Request-Method
< x-amz-bucket-region: us-east-1
< Server: AmazonS3

```

> 

"By default, CloudFront allows only the GET and HEAD methods, but some web browsers [i.e. Safari] might issue requests for the OPTIONS method."

"The CloudFront distribution's cache behavior allows the OPTIONS method for HTTP requests

If you're still seeing errors after you update your CORS policy and forward the appropriate headers, try allowing the OPTIONS HTTP method in your distribution's cache behavior. By default, CloudFront allows only the GET and HEAD methods, but some web browsers might issue requests for the OPTIONS method.

To enable the OPTIONS method on your CloudFront distribution, follow these steps:

1. Open your distribution from the CloudFront console.
2. Choose the Behaviors tab.
3. Choose Create Behavior, or choose an existing behavior, and then choose Edit.
4. For Allowed HTTP Methods, select GET, HEAD, OPTIONS."

https://aws.amazon.com/premiumsupport/knowledge-center/no-access-control-allow-origin-error/

PR #1701, opened 2020-09-14
Correct spelling of "woops" error message to more common usage

ripleymj:
Several people have pointed out that the error message with "woops" uses an unusual spelling, instead of the more common "whoops" as defined by Merriam-Webster. Attached is my somewhat amateur attempt to correct this, as well as the test cases. en-AU, en-CA, and en-GB localizations are left unmodified, as I cannot vouch for the local terminology.


simonista:
Hey @ripleymj , thanks for the PR, I'll get this pulled in.

simonista:
And it was correct not to modify the translation files, that will be left to our translators, although this will change the translation key (the disadvantages of auto-generated keys... but I can manage that with them)

simonista:
Merged in 501e653b717caee50bc03dbdb35767d7a9d9c615

PR #1712, opened 2020-10-01
sort classes in CoursesTray by displayed name

juniorRubyist:
This simple line in the React component for the CoursesTray ensures that the courses listed when you click the "Courses" tab gets sorted alphabetically, before being displayed to the user.


simonista:
Looks like this was fixed recently another way, with a1182ffe9b671baeaf9ddb64e180ebd4029ebab6. Thanks for the contribution!

PR #1718, opened 2020-10-07
Add federated attributes support for CAS logins

tilk:
This patch adds support for federated attributes to CAS authentication. It was tested at University of Wrocław.


ccutrer:
Can you add a spec showing that they're properly applied?

tilk:
I'm sorry, I'm not proficient in Ruby on Rails, I probably can't do it well without spending a lot of time on it. I looked at specs for OpenID and SAML and I don't know how to write this kind of test for CAS.

Still, this feature was required to make Canvas work for my organization, and [at least one other person wanted it, too](https://community.canvaslms.com/t5/Question-Forum/Does-CAS-support-federated-attributes-mapping-when-using/td-p/152198).

PR #1750, opened 2020-11-14
Exposed the trusted setting for developer keys

kylehovey:
Setting a key as trusted lets you skip the OAUTH approval screen when using the key. This is helpful for creating a key used for minting an OAUTH token for login with mobile apps (so that the users don't have to hit the approve button whenever they log in).


kylehovey:
This was meant for a fork.

PR #1758, opened 2020-11-28
.env as single source of truth for the entire app configs, migrate to mailhog instead of old outdated mail catcher 

MuhmdRaouf:
This PR contain multiple changes:
 - add .env.example to be used as .env to be a single source of truth for the entire app configs
 - migrate from mail catcher to mailhog for better modern UI, support, and faster mailer (build with GoLang)
 - add Makefile for easier accessibility instead of memorizing entire docker-compose commands 

so this PR may contain breaking changes but for development, It will make your life much easer just change anything you want in the .env file and it will reflect in ur app automatically after using 'make recreate' command to recreate the container with the new configs, also migrating to mailhog as it has better UI and faster mailer and added a volume with it to persist all the mails, and added a Makefile which containing docker-compose commands also it detect if the container is up and running or not and based on that it will run the appropriate command weather exec or run so it won't conflict with containers
use cases for that you don't have to run the entire application to run:
 -  Test suite `make test`
 -  Migration `make migrate`
 -  generating something using `rails generate` command `make generate [what you want] eg.: make generate controller something`


ccutrer:
Please break this PR up into discrete pieces that can be reviewed and submitted separately.

PR #1766, opened 2020-12-04
Grammar error in Norwegian under group discussion

manilpit:
Grammar error in Norwegian under group discussion


slaughter550:
@manilpit for Nynorsk - you can submit these changes here (https://www.transifex.com/instructure/canvas-crowdsource/) and for Bokmål - we have passed this change off to our partner. Thank you very much for your suggestions!

PR #1770, opened 2020-12-07
Fix for quiz timer issue #1769

### LINKED ISSUE #1769: Quiz Gets Auto-submitted Based On Upcoming Lock Time Instead Of Quiz Time Limit
### Summary:

The following code starts on line 674 of https://github.com/instructure/canvas-lms/blob/548a2a17327cd60a09b9b28513e7dbed2c5c56d5/app/models/quizzes/quiz.rb:

Link to the line: https://github.com/instructure/canvas-lms/blob/548a2a17327cd60a09b9b28513e7dbed2c5c56d5/app/models/quizzes/quiz.rb#L674

if !end_at || lock_at < end_at
    end_at = lock_at
end

The above logic results in an active quiz getting auto-submitted by Canvas based on the lock_at value, which can be earlier than end_at value if an instructor wants to restrict the starting time of the quiz, say, from 12PM to 12:30PM and the quiz's time limit is set to 90 minutes, which will result in the end_at time to be between 1:30PM and 2PM. In this case, end_at is set to a value in range 1:30PM and 2PM and lock_at is set to 12:30PM, as a result, lock_at < end_at evaluates to true and end_at is set to lock_at, which forces the quiz to be auto-submitted at 12:30PM for all students taking the exam causing a lot of frustration for students and instructors alike.

### Steps to reproduce:

 1. Create a quiz that students can start from time t1 to t2 (t1 < t2)(when time.Now -eq t2, lock the quiz)
 2. Set the time limit for the quiz to q1 where q1 + t1 > t2 (here, we have (q1 + t1) stand for the earliest time when the quiz can be submitted and t2 stands for the time at which the quiz gets locked preventing students to start a new quiz)
 3. Log in as a student and take the quiz from time t1 to t2, then see the quiz auto-complete at the time t2, which is less than the allotted time q1 for the quiz.

### Expected behavior:
Users should be able to set quiz's lock time and quiz's time limit independently of each other, such that if a user starts
the quiz before the lock time, the quiz will last at most for duration of the time limit set for the quiz.
### Actual behavior:

The lock time overwrites the quiz's time limit.

### Additional notes:
Please fix it sooner than later as it affects large groups of students during high-stress times like exams and test.

abbottjord94:
Implements a fix for Issue #1769, where quiz timers would default to the latest availability time of a quiz, instead of granting the full time limit as set by the instructor.

I was able to recreate the issue as described. The fix was very simple: I simply set `lock_at `to equal `end_at `when there is no `end_at` set, or when `lock_at < end_at`. This means that the student should be granted the full time limit, regardless of when they start the quiz within the allotted Available from/Until times.

closes #1769 

Test Plan:
- Create a quiz with an Available from/Until time with a short range (I chose 1 minute)
- Set the Time Limit of the quiz to something greater than the Available from/Until range (I chose 5 minutes)
- Verify that the quiz timer grants the allotted Time Limit to the student for that attempt.

### DIFF
--- a/app/models/quizzes/quiz.rb
+++ b/app/models/quizzes/quiz.rb
@@ -672,7 +672,7 @@ class Quizzes::Quiz < ActiveRecord::Base
     # set to lock date
     if lock_at && !submission.manually_unlocked
       if !end_at || lock_at < end_at
-        end_at = lock_at
+        lock_at = end_at
       end
     elsif !end_at || (fallback_end_at && fallback_end_at < end_at)
       end_at = fallback_end_at


slaughter550:
Hello and thank you for your submission. While we can see that this change does update according to the use case you have described. Unfortunately the change is in conflict with the originally designed classic quiz behavior that the community has come to rely on. We recently added the lock at end date to New Quizzes because there was a substantial demand for educators to continue seeing this behavior. We appreciate your dedication to making Canvas better and hope this feedback doesn’t discourage you from future suggested changes.

PR #1784, opened 2020-12-23
Removed media query to hide assignment availability

gewethor:
Removed media query that hides assignment availability within the home tab on screens smaller than 1280px. There appears to be no reason the assignment availability should be hidden and hiding the availability removes the information for users on smaller screens and creates an excess of whitespace.


slaughter550:
@gewethor your pull request has been merged in. Thanks so much for your contribution!

PR #1803, opened 2021-01-29
Support for running without Upstart or Dory

adamcrown:
These changes make the `docker_dev_setup.sh` able to be used alongside our MasteryConnect Docker setup, specifically, and should also make it more compatible with other OSes and dev environments in general.

**Changes:**

If the `service` command does not exist, fall back on `systemctl`.

Allow a `SKIP_DORY` ENV var to be set that will bypass installing and
starting Dory for compatibility with other tools such as
`letsencrypt-nginx-proxy-companion`.

**Test Plan:**

- On a dev box, you should be able to run the `script/docker_dev_setup.sh` command and it should work as it normally does on a regular Canvas dev box.

- On an OS without a `service` command, and the `docker` service stopped, you should still be prompted to start the docker daemon, and it should start successfully.

- If you run `script/docker_dev_setup.sh SKIP_DORY=true` you should not see `Starting dory...` anywhere in the output of the command.


adamcrown:
Closing this because it's getting implemented a bit differently through an internal PR.

PR #1810, opened 2021-02-08
Show file size depend on the OS

Gao-Jun:
One day, a Windows user asked, why I got a smaller file downloaded (806K) compared with the size shown in 'Files'(825K) ? Is the file downloaded completely?

In Mac OS, 1 KB = 1000 Bytes. In Windows and Linux, 1 MB = 1024 Bytes

This patch will display the file size depend on the OS.

In detail, 1 MB = 1000 Bytes for Mac user, and 1MiB = 1024 Bytes for others

### DIFF
--- a/app/coffeescripts/util/friendlyBytes.js
+++ b/app/coffeescripts/util/friendlyBytes.js
@@ -17,19 +17,26 @@
 
 import I18n from 'i18nObj'
 
-const units = ['byte', 'bytes', 'KB', 'MB', 'GB', 'TB']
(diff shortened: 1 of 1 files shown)


dustin-cowles:
Hi @Gao-Jun Sorry this has not been addressed in some time. We are working on getting to things much quicker now. This file has been moved and thus this PR is no longer valid. If you wanted to update it I could work on getting it reviewed.

Gao-Jun:
Hi @dustin-cowles An updated version of this PR is here #2367

PR #1812, opened 2021-02-09
Add 'host' and 'username' entry to 'development' database.yml.example

Gao-Jun:
In `config/database.yml.example` , both `production` and `test` have 7 config entries, while `development` only have 5 with `host` and `username` missing.

This PR add these two config entries to the sample config file.

### DIFF
--- a/config/database.yml.example
+++ b/config/database.yml.example
@@ -12,6 +12,8 @@ development:
   adapter: postgresql
   encoding: utf8
   database: canvas_development
+  host: localhost
+  username: canvas
   password: your_password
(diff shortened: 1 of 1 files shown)


ccutrer:
this is intentional. it's assumed for development you'll be using a local postgres, with trust authentication over a unix socket. if you use `host` that will make it use TCP, not UDP. by omitting `username` postgres will use your current username, which is likely what you want. if you need something different than that, you're free to modify your local database.yml - this _is_ just an example, but happens to be the most common configuration.

PR #1839, opened 2021-04-01
Revert course title alignment in classic planner

rjacoby00:


spencerolson:
Hi @rjacoby00  , thanks for the contribution. We made this alignment change intentionally but are open to changing it back with enough support from the community. To propose this idea, please create a new "Idea Conversation" (https://community.canvaslms.com/t5/Community/How-do-I-create-a-new-idea-conversation-in-the-Canvas-Community/ta-p/2986) and we'll keep an eye on the support level. I'm going to close this for now, and we can re-open if we decide to make this change. Thanks!

PR #1847, opened 2021-04-10
Removes some harmful language in specs

rthbound:
The primary purpose of this change is to eliminate a paradigm of "student is slave and teacher is boss".

Other small changes:
- Worthless loser?
- Soften a somewhat disparaging comment regarding a perceived code smell

### DIFF
--- a/spec/models/course_spec.rb
+++ b/spec/models/course_spec.rb
@@ -1569,9 +1569,9 @@ describe Course do
       expect(@course.course_section_visibility(unlimited_teacher)).to eq :all
     end
 
-    it "returns none for a nobody" do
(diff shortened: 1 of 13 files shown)


simonista:
Hey @rthbound , thanks for making this change, this is great. I'll get it run through our CI pipeline today.

rthbound:
Thank you @simonista -- I was hoping the change would be accepted.

simonista:
Just a quick update here, we have a campsite rule that checks all specs touched by a commit, and two of the specs touched here are sometimes too slow, so I'm looking into how to speed them up.

rthbound:
@simonista which two? maybe I could have a look. Is there a chance these two slow ones may be covered elsewhere (and could be removed)?

PR #1869, opened 2021-05-05
Update grammer for enrolling in course

manilpit:
Just a simple grammer update. We frequently get supportcases based on little errors like this and hope that we can get through with some small grammer changes. Thank you.

### DIFF
--- a/config/locales/nb-x-k12.yml
+++ b/config/locales/nb-x-k12.yml
@@ -18756,7 +18756,7 @@ nb-x-k12:
         go_to_dashboard: "Gå til oversikten din"
     authenticate: 
       buttons: 
-        enroll_in_course: "Registrere deg på fag"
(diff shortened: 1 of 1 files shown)


spencerolson:
Hi @manilpit , thanks for your contribution! 

For changes to our translation files, we work with a 3rd party, meaning we don't manually modify or make changes to those files directly. However, I had your suggested changes reviewed, and we’ve decided to let the 3rd party know that we'd like the translations changed to `enroll_in_course: "Registrerer deg i fag"` and  `getting_started: "Du registrerer deg i **%{course}**"` 

Thanks again!

manilpit:
Thank you for the feedback. Since our userbase often find these little grammatical changes i am left wondering who the 3rd party might be, and if one could open a direct dialog with them to create a more efficent dialog. Thank you again.

PR #1875, opened 2021-05-20
Use SHA256 for cache keys

mpetrowi:
This PR updates cache key computation to use SHA256 instead of MD5.

This change should not affect site operation, so the running the specs should be a sufficient test.

Test plan:
- specs pass


dustin-cowles:
@mpetrowi Most or all of these changes have already been merged over the last few years. While we appreciate your contribution, this PR is not needed at this time. I did see some remaining MD5 usages elsewhere in the code so perhaps there is still work to be done here. Feel free to open another PR if you feel inclined.

PR #1877, opened 2021-05-20
Use SHA256 uuids during importing

mpetrowi:
This PR changes the digest used to generate import UUIDS from MD5 to SHA256.

Test plan:
  - specs pass
  - regression test course copies and Canvas cartridge imports
  - regression test blueprint copies


dustin-cowles:
@mpetrowi Like the last couple of PRs, if you wanted to update this I could pull it into our CR pipeline. Sorry for the delay!

dustin-cowles:

PR #1880, opened 2021-05-20
Use SHA256 for due_date_cacher job strand name

mpetrowi:
This PR changes the due_date_catcher job strand name to be based SHA256.

Test plan:
- spec spec/lib/due_date_cacher_spec.rb passes


simonista:
Merged in a05dc0f8dd6d9a89369bae7828cffd0f1cb2631d

PR #1883, opened 2021-05-20
Use SHA256 for UUID on external feed items

mpetrowi:
This PR changes the algorithm used to generate external feed UUIDs from MD5 to SHA256.

In some cases external feed entries can be missing the guid or have a guid
which is too long. Previously we were using an MD5 to generate a uuid
on our end. This change switches to using SHA256.

For existing entries, we fall back to matching on the entry link, so in most
cases this change should not result in duplicate feed entries being
added during the transition. Feeds that match based on the entry link will get
updated with the new uuid.

Test plan
- regression test rss and atom feeds


simonista:
Merged in a209dba242bcc3a7b28d86b3fe17c182c22b4b91

PR #1884, opened 2021-05-20
Add option to use SHA512 for stored file digests

mpetrowi:
This PR adds the ability to use SHA512 for locally stored file digests.  This change only affects local file storage, and the default remains MD5.

Configuration for the digest is done using a new 'use_sha512_digests' setting
in config/file_store.yml.  The allowed values are:
- false: continue to use md5 sums  (default)
- true: always use SHA512 digests for new files

Migration: When changing `use_sha512_digests` to true, existing files
and attachments will continue to function as before and new uploads will
use SHA512. This means that new uploads will not match existing files
for deduplciation purposes until the existing MD5s are recomputed.

To recompute hashes on existing attachments, run the rake task:
  rake db:migrate_attachments_to_sha512

Implementation: We're using the md5 column in the attachments table to store the SHA512
hash. This isn't ideal, but instfs SHA512 file hashes are already stored in
that column and s3 uses etags in that column. The column could be renamed to `digest`
in a future commit.

Test plan:
- all tests should be run with local file storage
- spec/models/attachment_spec.rb, spec/lib/unzip_attachment_spec.rb pass
- spec/lib/data_fixup/migrate_attachment_digests_spec.rb passes
- using the default file_store config with local file storage:
  - regression test file uploads
  - regression test uploading a zip file and unzipping
  - regression test uploading images in the RCE
  - regression test thumbnails for uploaded images
- add `use_sha512_digests: true` to config/file_store.yml and repeat the
above regression tests
- migration test:
  - remove `use_sha512_digests` from config/file_store.yml and restart
  rails
  - upload the following new files (never before uploaded to this instance):
    - an image
    - a non-empty zip file and have canvas expand it
  - add `use_sha512_digests: true` to config/file_store.yml and restart
  rails
  - run `rake db:migrate_attachments_to_sha512` and wait for all the
    DataFixup:MigrateAttachmentDigests jobs to finish
  - verify that the new uploaded files still work
  - verify in the db that the uploaded attachments contain a sha512 digest
  in the md5 column
  - upload the same image and zip file to a new course
  - verify in the db that the new attachments have the root_attachment_id
  field set to the previously uploaded files

### DIFF
--- a/app/models/attachment.rb
+++ b/app/models/attachment.rb
@@ -142,7 +142,8 @@ class Attachment < ActiveRecord::Base
       :path_prefix => file_store_config['path_prefix'],
       :s3_access => 'private',
       :thumbnails => { :thumb => '128x128' },
-      :thumbnail_class => 'Thumbnail'
+      :thumbnail_class => 'Thumbnail',
+      :use_sha512_digests => file_store_config['use_sha512_digests'],
   )
 
   # These callbacks happen after the attachment data is saved to disk/s3, or
--- a/app/models/thumbnail.rb
+++ b/app/models/thumbnail.rb
@@ -31,7 +31,8 @@ class Thumbnail < ActiveRecord::Base
       :path_prefix => Attachment.file_store_config['path_prefix'],
       :s3_access => 'private',
       :keep_profile => true,
-      :thumbnail_max_image_size_pixels => Setting.get('thumbnail_max_image_size_pixels', 100_000_000).to_i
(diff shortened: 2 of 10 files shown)


dustin-cowles:
@mpetrowi I am going to label this entire series as waiting for contributor. If you still with to get these merged get them updated and let me know. Sorry for the (incredibly long) delay. If you no longer wish to merge them go ahead and close them or I will in some time.

dustin-cowles:

PR #1885, opened 2021-05-20
Use SHA256 hash for branding config

mpetrowi:
This PR changes the brand configs to use SHA256 rather than MD5 as the unique identifier for new configurations. It should not break backwards compatibility with existing brand configs stored by MD5.  There could definitely be improvements to clean up the PR (renaming db columns, etc) although I wanted to elicit some feedback before making additional changes.

Implementation notes:
- After the change, existing brand configs will continue to be stored by
md5 and they will continue to operate as before. The brand configs are
designed to be immutable, so updating the existing hashes is probably not
a good idea.
- The migrations which create the predefined themes now use static MD5s
for compatability reasons.
- The database columns continue to be named md5, which isn't ideal.
These could be renamed to 'digest' when this is incorporated upstream
which would mean renaming fields in multiple models.

Test plan
- reset the database and run migrations. the default themes should work
as before
- regression test theme configuration and saving


db/migrate/20150709205405_create_k12_theme.rb
```diff
@@ -33,7 +34,11 @@ def up
     bc = BrandConfig.new(variables: variables)
     bc.name = NAME
     bc.share = true
+    bc.md5 = MD5_OF_K12_CONFIG
+    # set the md5 directly in the record for backwards compatability
+    BrandConfig.skip_callback :validation, :before, :generate_md5
```
ccutrer, line 39: you should use suspend_callbacks instead

ui/shared/theme-editor/react/PropTypes.js
```diff
@@ -19,7 +19,7 @@
 import PropTypes from 'prop-types'
 import _ from 'underscore'
 
-const MD5_REGEX = /[0-9a-fA-F]{32}$/
+const MD5_REGEX = /[0-9a-fA-F]{32,255}$/
```
ccutrer, line 22: why 255? shouldn't it be exactly 32, or exactly 64?

mpetrowi: This is fixed.  I also added a '^' anchor.

db/migrate/20150709205405_create_k12_theme.rb
mpetrowi: That's a good way to do it.  I pushed a change, but it's not right.  I'll get this fixed

mpetrowi: Should be fixed now.

mpetrowi:
Hi Cody,  One more change: I updated the md5 model validations to match what's check client-side.  I'm amending a single commit for these changes, I hope that's okay on your end.  Thanks!

dustin-cowles:
@mpetrowi I am going to label this entire series as waiting for contributor. If you still with to get these merged get them updated and let me know. Sorry for the (incredibly long) delay. If you no longer wish to merge them go ahead and close them or I will in some time.

dustin-cowles:

PR #1895, opened 2021-06-09
Adding a space after self-enrolled URL, before period

bsnyder9:
Course creators occasionally accidentally include the period when copy/pasting the self-enroll URL, and less tech-savvy students may not know to remove it

### DIFF
--- a/app/views/courses/settings.html.erb
+++ b/app/views/courses/settings.html.erb
@@ -451,7 +451,7 @@
               <%= t 'course_open_enrollment', <<-TEXT, :url => enroll_url(@context.self_enrollment_code || '{{ self_enrollment_code }}'), :url2 => register_url, :code => @context.self_enrollment_code || '{{ self_enrollment_code }}', :wrapper => '<b>\1</b>'
                     This course has enabled open enrollment. Students can
                     self-enroll in the course once you share with them this URL:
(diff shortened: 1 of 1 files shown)


bsnyder9:
@slaughter550 Any chance this request could get a look? The issue keeps coming up in trainings for users in my district.

app/views/courses/settings.html.erb
```diff
@@ -451,7 +451,7 @@
               <%= t 'course_open_enrollment', <<-TEXT, :url => enroll_url(@context.self_enrollment_code || '{{ self_enrollment_code }}'), :url2 => register_url, :code => @context.self_enrollment_code || '{{ self_enrollment_code }}', :wrapper => '<b>\1</b>'
                     This course has enabled open enrollment. Students can
                     self-enroll in the course once you share with them this URL:
-                    *%{url}*. Alternatively, they can sign up at *%{url2}* and
+                    *%{url}* . Alternatively, they can sign up at *%{url2}* and
```
slaughter550, line 454: @bsnyder9 I would just change this to read as `them this URL: *%{url}* or they can sign up at` and leave the punctuation issue out of it

slaughter550 (changes requested):

app/views/courses/settings.html.erb
bsnyder9: Makes sense to me!

PR #1909, opened 2021-07-02
Fix example for lti.showModuleNavigation in docs

buckett:
Subject for postMessage was incorrect in the example for lti.showModuleNavigation


buckett:
@xandroxygen This looks to just be a copy/paste error.

PR #1914, opened 2021-07-10
Updated the BBB plugin to accept any API URL

ianespana:
Some BBB service providers have configured their API URLs differently than the default `https://bigbluebutton.mycompany.com/bigbluebutton/api/`.

For instance, a service provider called MynaParrot utilizes `https://api.mynaparrot.com/bigbluebutton/username/api/` as their API URL. This would not work with Canvas as it was expecting the URLs to be formatted like the default one, and thus limited the amount of services it'd work with quite severly.

Now we just make sure said URL ends with `/api/` instead of assuming the default format is being used.


maths22:
Thank you for your contribution. While I like the idea of this change, as currently implemented this will break any existing plugin configurations.  I would suggest either adding a migration to automatically transform existing plugin settings to use the new format (if you do that you'll need to migrate into a new field name in a predeploy migration (e.g. :address instead of :host) so that conferences still work after the migration has run before the new code is deployed) or I would suggest detecting if the url starts with `http(s)://` and using the new logic if it does and the old logic if it doesn't.

Also please don't update the language .yml files, those will be automatically updated separately through an automatic process.

ianespana:
Oh, right. I'll implement the second way (check if https:// exists). As for the translations I did not know, so I will revert those

ianespana:
Changes done, it now uses the old logic if http(s):// is not set in the domain configuration

dustin-cowles:
Hi @ianespana , sorry this PR got missed. If you can verify it is still valid and resolve the conflicts, I can get it reviewed. Thanks!

dustin-cowles:
I am going to go ahead and close this PR. If it is still valid please update and reopen.

Issue #1856, opened 2021-04-19
Log rotate does not work as expected with rails logger


torebre:
Summary: Log rotate does not work as expected when using the rails-logger.

Steps to reproduce: Use rails logger and set debug level: The log does not rotate. Set the the log level to something else and the log rotates.

Expected behavior: The log level should not affect the log rotation.

Additional notes: It looks like the level argument is being given as input where the super class expects the shift_age argument to be: 
https://github.com/instructure/canvas-lms/blob/181cab70ed355709d23bff303fd70764d4305486/lib/canvas_logger.rb#L29

torebre:
This is no longer an issue, canvas_logger.rb has been updated since this issue was opened.

PR #1928, opened 2021-08-21
Setting level-argument for logger, it was earlier set in the position…

torebre:
… where the shift-argument is expected in the logger.

This commit will cause the logs using the rails logger to no longer rotate no matter what the log level is. Previously the logs would not rotate only if the level was set to debug, because in this case the shift-argument was set to 0. For log level info for example the shift argument would be set to 1, causing the logs to rotate.

Closes gh-1856

Test plan:
- Specify "logger: rails" and "log_level: info" in logging.yml
- Observe that the logs will grow beyond 1 MB


ccutrer:
CanvasLogger no longer overrides `initialize`, so this should no longer be a problem.

PR #1932, opened 2021-09-03
Restrict join URLs to a conference

bfcoder:
When a BigBlueButton conference is created and active, this current conference has a createTime associated with it.  This pull request adds the createTime parameter to join URLs so the URLs are only valid for the current conference.


roor0 (commented):
Generally we store timestamps as timestamps instead of integers. We then convert to the needed format where it is used or consumed. That being said, it looks like we just want to store a common key/integer, is there a reason we cannot just put this into the conference settings with a new_key instead of adding a new column?
I think putting it into settings simplifies the change, and makes it not add a column for all conference types to only be used by BBB

bfcoder:
@roor0 Thanks for the heads up about the conference settings. I didn't realize that was a thing. I've removed the migration in favor of the conference settings. Look better now?

As for how it is stored, the integer for the epoch is just what BBB sends back. I don't have any control over the data type. And I don't think it makes sense to convert it to a timestamp just to convert it back to an epoch later. So it makes sense just to store the epoch.

PR #1933, opened 2021-09-03
Add ability for LTI tools to send toast messages

claydiffrient:
This adds the ability for LTI tools to send toast alert messages through the
Canvas alert system.  This enables them to get the benefit of having a
fully integrated alerting solution including screen reader
announcements.  It also enables messages to be visible regardless of
scroll location within the external tool's iframe.

To facilitate ease of knowing where the message is coming from, the name
of the external tool is prepended to the message sent by the tool.

Test Plan:
  - Have an LTI tool send a "lti.alertMessage" message
  - The alert should show as "Tool Name: Alert Message Text"
  
This is what the alert looks like: 
Screen Shot 2021-09-03 at 2 00 25 PM

### DIFF
--- a/app/models/lti/launch.rb
+++ b/app/models/lti/launch.rb
@@ -62,6 +62,10 @@ module Lti
       url.path.presence || '/'
     end
 
+    def tool_name(context)
+      ContextExternalTool.find_external_tool(analytics_id, context).name
+    end
+
     def analytics_id
       @analytics_id || URI.parse(resource_url).host || 'unknown'
     end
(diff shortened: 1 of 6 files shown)


simonista:
Hey @claydiffrient , thanks for the PR. I'll make sure interop knows about it.

claydiffrient:
Looks like a modified version of this was merged via c3760f773cbde4a74c7be385c0b1567666fd2d8e so I'm closing this one out.  Thank you so much!

PR #1941, opened 2021-09-21
Add setting to change TTL on AccessVerifier

bfcoder:
5 minutes sometimes just isn't enough. This makes it so it is configurable and not hard coded.

Test Plan:

* Add a video to a course
* insert a video tag into the html of a page. ie:
```
<p><video style="border: 2px solid black;" width="450" height="225" controls="controls">
<source src="/courses/<course_id>/files/<file_id>/download" /></video></p>
```
* Inspect the network traffic while watching the video.
* Obtain the jwt used to download the video.
* Inspect that jwt and the exp should be the minutes out of what ever is set in the Setting, default 5 minutes.


bfcoder:
@maths22 Any chance you can take a look at this soon? Thanks!

PR #1943, opened 2021-09-27
Update data service event ( submission_updated ) documentation

johnsalzy:
## Summary of the commit

Change the '|' to a ',' as it was attempting to make a new table column, causing documentation data to be cut off.

#### Documentation Before Fix ( Web Preview )
documentation before fix web preview

#### Documentation Before Fix ( Rendered Markdown Preview )
documentation before fix

#### Documentation After Fix ( Rendered Markdown Preview )
documentation after fix

## Test Plan:
 - Pull code
 - Render new markdown file and ensure new output is correct
 - Render it for the web preview

### DIFF
--- a/doc/api/data_services/md/dynamic/data_service_canvas_submission.md
+++ b/doc/api/data_services/md/dynamic/data_service_canvas_submission.md
@@ -231,7 +231,7 @@ Submission
 | **missing** | Whether the submission is missing, which generally means past-due and not yet submitted. |
 | **score** | The raw score |
 | **submission_id** | The Canvas id of the new submission. |
-| **submission_type** | The types of submission ex: ('online_text_entry'|'online_url'|'online_upload'|'media_recording') |
+| **submission_type** | The types of submission ex: ('online_text_entry','online_url','online_upload','media_recording') |
 | **submitted_at** | The timestamp when the assignment was submitted. |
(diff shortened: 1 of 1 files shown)


spencerolson (changes requested):
Hi @johnsalzy thanks for catching and fixing this!

So the way that this portion of our API docs are generated are:
1. A change is made to the `.json` file (in this case, `doc/api/data_services/json/canvas/event-types/submission_updated.json`)
2. We run `bundle exec rails doc:api` to generate the markdown (in this case, `doc/api/data_services/md/dynamic/data_service_canvas_submission.md`)
3. The resulting changes are committed

The way the commit is as-is, if we run `bundle exec rails doc:api`, the change will be reverted in the generated markdown file because the change was not applied to the `.json` file. So, you just need to apply the same change (replacing `|` with `,`) in `doc/api/data_services/json/canvas/event-types/submission_updated.json`. Let me know if you have any questions, thanks!

scoutskylar:
I made the requested changes at https://github.com/instructure/canvas-lms/commit/738060a88489d8a8024d6f9f291b1f465c991c9f

scoutskylar:
I made the requested changes in a new pull request, which does not need conflicts resolved: https://github.com/instructure/canvas-lms/pull/2011

PR #1979, opened 2021-11-18
Remove redis `auth` from unsupported commands

grahamb:
Canvas lists a number of Redis commands that are (or were) unsupported by twemproxy and prevents them from being used outside specs or rails console. This includes the `auth` command, which means that Canvas can not connect to a redis instance using password authentication.

Via @maths22 in IRC, Canvas no longer uses twemproxy, so the restriction on `auth` can be removed.

Screen Shot 2021-11-18 at 11 54 24


maths22:
Fixed in #e4b61c84e700c84d4194135a5ec04fb5deb38751

PR #1980, opened 2021-11-22
Add points_possible to censored quiz_question API

ndrewh:
Currently the API does not return points_possible for questions from the quiz_question API when returning a "censored" response. Since currently Graders get censored responses from this API, this makes it very difficult to grade questions via the API.


ndrewh:
🎂 Any update on this? The points_possible field is displayed on the web to students/graders, but still not in the API.

PR #1457, opened 2019-05-28
Allow account admin to modify the state of profile pic via API

Gao-Jun:
**Summary**
This patch allow account admin to 'lock' the profile pic via API.

**Background**
Currently, account admin cannot lock a profile pic via API. So, it's very difficult for admins to use 'official' photos and forbid users to  change it.

Here is the logic of profile pic:

1. ’Enable Profiles' not checked in account settings
Profile pic only shown on the top of menu.

2. 'Enable Profiles' checked and the 'avatar_state' of user is not 'locked'
Profile pic shows in multiple tools, and user can edit their own pic.

3. 'Enable Profiles' checked and the 'avatar_state' of user is 'locked'
Profile pic shows in multiple tools, and user can not edit their own pic.

So, if an account admin doesn't want users to use self-uploaded pic, it's better to 'lock' the avatar_state. Yet, it seems no API support this function. In the community, some people has to override users pic every day. See [https://community.canvaslms.com/thread/14576-show-official-photo-not-avatar-in-attendance](url)
This solution cannot forbid users upload their own pic, and their may fell strange that their pic will be override in the next day.

Or, use customized CSS and JS to hide the edit function. See[https://community.canvaslms.com/thread/25996-how-can-i-hide-with-css-the-ability-for-users-to-click-their-profile-avatar-in-order-to-change-it](url)
Users still can upload their own pic with some tricks.

**Solution**
This patch will allow account admins to 'lock' the pic and solve this problem.
To achieve this goal, a new parameter **user[avatar][state]** is added to 

> PUT /api/v1/users/:id

An account admin can set the state to 'locked' to lock the profile pic. Even more, an account admin can lock some users' profile pic and allow others to customize it.

**P.S.**
I also find in 'route.rb', there is a route can modify the avatar_state directly.
```ruby
put 'images/users/:user_id' => 'users#update_avatar_image', as: :update_avatar_image
```
I failed to find where to invoke this, and this route is added in the at least 5 years ago.
This route doesn't support invoke by Restful API. 
I think add Restful API support to it can also achieve the same goal.

PR #1993, opened 2021-12-28
Allow directly changing avatar state through API

dc74089:
### Allow directly changing avatar state through API

This allows for Canvas admins at institutions to lock, unlock, and approve user avatars in bulk, as has been requested in the Canvas community forum ([thread A](https://community.canvaslms.com/t5/Canvas-Question-Forum/Adding-approving-and-locking-avatars-via-the-API/td-p/49765), [thread B](https://community.canvaslms.com/t5/Idea-Conversations/Mass-upload-Avatars-and-then-have-them-locked-but-viewable/idi-p/423883)). 

In adding code to the users controller, I followed the format and order of similar functions the best I could. I have intentionally ignored the linter at `app/controllers/users_controller.rb:2035 => Style/HashLikeCase` as I believe the suggested format sacrifices readability in a very big way. I am happy to modify the code if Instructure would prefer to prioritize the slight performance boost of using a hash. 

### Test Plan
In an environment with admin and non-admin users already created, execute the following:

1. `PUT http://{canvas_url}/api/v1/users/{user_id}` with data `user[avatar][state]=locked` and an admin's API token
- Observe that the avatar state is changed in the database
- Observe that user `{user_id}` is unable to change their profile photo using the UI
2. `PUT http://{canvas_url}/api/v1/users/{user_id}` with data `user[avatar][state]=unlocked` and a non-privileged user's API token
- Observe that the avatar state is not affected, as non-privileged users cannot modify their own avatar state
3. `PUT http://{canvas_url}/api/v1/users/{user_id}` with data `user[avatar][token]={a valid token}` and and admin's API token
- Observe that an admin can always change a user's avatar, even when it is locked
- Observe that the avatar state remains locked
4. `PUT http://{canvas_url}/api/v1/users/{user_id}` with data `user[avatar][state]=unlocked` and an admin's API token
- Observe that the avatar state is changed in the database
- Observe that user `{user_id}` is again able to change their profile photo using the UI

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -1949,6 +1949,10 @@ class UsersController < ApplicationController
   #   token and instead pass the url here. Warning: For maximum compatibility,
   #   please use 128 px square images.
   #
+  # @argument user[avatar][state] [String]
+  #   The desired lock state of a user's avatar. Allowed values include:
+  #   ['none', 'submitted', 'approved', 'locked', 'reported', 're_reported']
+  #
   # @argument user[title] [String]
   #   Sets a title on the user profile. (See {api:ProfileController#settings Get user profile}.)
   #   Profiles must be enabled on the root account.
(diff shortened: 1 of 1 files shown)


dc74089:
Upon further review, this appears to be a functionally equivalent duplicate of #1457, which has gone stale. I will keep this open as it will likely be easier to merge my code if accepted, since it is written on top of a much more current fork.

Gao-Jun:
As the author of #1457, I have to say it takes too long time to let Instructure accept a PR.

dc74089:
@maths22 I just accepted the CLA. What are the next steps of getting my code reviewed and implemented?

maths22:
#1457 has been merged which implements this functionality.

PR #2009, opened 2022-01-28
Document the skip_message parameter.

buckett:
All reports support the `skip_message` parameter which when set prevents Canvas from sending the user a notification about the completion of the report, this should be in the account report documentation.


maths22:
Merged in 03d2679c58081f9c11a1b6a2a46d8b4ada26f72f

PR #2011, opened 2022-02-01
Change list format to match rest of docs

scoutskylar:
Most lists like this in the docs use commas. This pull request changes the vertical pipes to commas and removes the single quotes around each list item. Using vertical bars causes the text to not render properly, as explained in https://github.com/instructure/canvas-lms/pull/1943.
Closes https://github.com/instructure/canvas-lms/pull/1943

### DIFF
--- a/doc/api/data_services/json/canvas/event-types/submission_updated.json
+++ b/doc/api/data_services/json/canvas/event-types/submission_updated.json
@@ -15,7 +15,7 @@
     "missing": "Whether the submission is missing, which generally means past-due and not yet submitted.",
     "score": "The raw score",
     "submission_id": "The Canvas id of the new submission.",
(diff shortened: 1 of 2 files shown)


spencerolson (changes requested):
Hi @scoutskylar so we actually need to commit changes to _both_ the `submission_updated.json` file as well as the `data_service_canvas_submission.md` file. If you run:

```
bundle exec rails doc:api
```

you should see that the `data_service_canvas_submission.md` has been modified, and then you can commit that modification along with the one you have here. Let me know if you have any questions and thank you for this pull request!

scoutskylar:
I don't know how to use Rails, but I edited the file `data_service_canvas_submission.md` manually. Will that work @spencerolson?

spencerolson:
@scoutskylar I think your change looks good, I'll double-check locally to make sure your change to the markdown file matches what would have been generated from the command I shared.

PR #2017, opened 2022-02-16
Clear out the section name input box after creating a new section

zwily:
In the Course Settings -> Section tab, when you added a new
section, the input box would remain with the newly created
section name. The code was apparently attempting to clear it
but needed an empty string to actually succeed.

Test Plan:
 - In a Course Settings -> Section tab, add a new Section
 - Behold that the input box is cleared after adding


ccutrer:
You should probably write a test for this.

maths22:
Merged in d1c856cde3b6bab60b5074d93431779bd915a322

Issue #1997, opened 2022-01-07
./script/docker_dev_setup.sh runs failed by docker doesn't support docker-machine anymore


huangdengpan:
the quick start introduction looks expired, how can i run canvas as dev environment?

jedieaston:
The quick start introduction (https://github.com/instructure/canvas-lms/wiki/Quick-Start) doesn't reference Docker Machine anymore. If you are on macOS, Windows (running WSL2 w/ Ubuntu and Docker installed), or Ubuntu those instructions should work.

afeld:
Just ran into the same issue; the script seems to be checking for it here:

https://github.com/instructure/canvas-lms/blob/4e0a9e39debe8d39df7a1d5dfdad72f2826b58f5/script/common/os/mac/dev_setup.sh#L39

huangdengpan:
Thank you all, i will try.

ccutrer:
Please use [the mailing list](https://groups.google.com/g/canvas-lms-users) for installation and operational issues.

PR #2018, opened 2022-02-16
fix use of Dory for Mac setup

afeld:
This fixes a couple issues:

- The Mac setup script referenced docker-machine, which hasn't been supported since 2019.
- If not using Mutagen, the setup script required Dinghy instead of allowing for Dory.

Since Dinghy also hasn't been maintained since 2019, I'd personally recommend dropping all references to that, but I figured that would be more controversial so didn't bother.

Fixes https://github.com/instructure/canvas-lms/issues/1997.

## Test plan

- Follow the [automated setup
  instructions](https://github.com/instructure/canvas-lms/wiki/Quick-Start#automated-setup)


maths22:
As of https://github.com/instructure/canvas-lms/commit/08965f67cadf29014f4c60958c013ad7ceffbea7 , the setup scripts have been updated to remove dinghy

PR #2031, opened 2022-03-16
Expose External Tool deployment_id in REST API.

tim-schilling:
It should be possible to access the deployment_id propery
when creating external tools via the REST API.

Test Plan:
 - Run external tool spec tests.
 - Create an external tool using a developer key and verify
   the deployment_id property is included in the response.


tim-schilling:
Closed in 9d59a2f3a1aab019fd574e905da0b87aaef87dde

PR #2032, opened 2022-03-17
Fix minor issues with locales.yml

KiwiBird97:
Here are the fixes that I am proposing. Note that these are only minor changes so that we don’t want to break any critical functionality in Canvas.
1. es_ES exists twice. Because of this, I looked up some strings on the Internet to determine which one is Castilian Spanish and which one is Latin American Spanish. The phrase “a las” is used for time phrases in Castilian Spanish (commonly known as Spanish (Spain)), so we left Español (España) untouched and changed the other from Español to Español (Latinoamérica).
(The bigeasy_locale is unchanged to avoid breaking critical functionality.)
2. English (United Kingdom) and English (US) do not appear to be in line. I chose English (US) to be replaced with English (United States).
3. Some languages are misspelled or not capitalized. We fixed them just as they appeared in the language menu of smartphones and tablets (placed in the following order: Latin, Cyrillic, right-to-left, languages, Indian, and East Asian). The variant is shown in parentheses only if a language has two or more regions/variants, just like in many mobile operating systems.

### DIFF
--- a/config/locales/locales.yml
+++ b/config/locales/locales.yml
@@ -139,7 +139,7 @@ da:
       weekday: "%A"
   dow_offset: 1
   locales:
-    da: Dansk
+    da: Dansk (Danmark)
   time:
     event: "%{date} kl. %{time}"
     formats:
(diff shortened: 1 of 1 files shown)


wutao007:
Initialization asset reported wrong, is it the code you submitted?

PR #1986, opened 2021-12-12
Add devcontainer configuration for GitHub Codespaces.

jedieaston:
This commit adds a devcontainer.json file and some associated configuration files to automatically configure a Canvas development environment in GitHub Codespaces (or locally, via the VSCode remote containers extension). This way, all you have to do is click "New Codespace" (or open the repo in VSCode locally), wait ten minutes, and be ready to contribute!

Test Plan:
- Create a new Codespace using a branch with these files. (Or open it in VSCode locally, and follow the prompts to build a devcontainer.)
- Make sure it builds successfully.
- Run "bundle exec rails db:initial_setup" and "bundle exec rails server" to make sure Canvas comes up successfully.

PR #2038, opened 2022-03-25
[CODESPACES] Adds configuration for starting app in codespaces

octosteve:
Hello! I wanted to contribute a configuration for getting up and running with Codespaces. It closely modeled after [this setup](https://github.com/microsoft/vscode-dev-containers/tree/main/containers/ruby-rails-postgres).

The resulting environment is one with gems and assets installed, and an admin user. You are also opted out of stats collection since this part is not done interactively.

Account name: learn_good_things_school
Admin Email: admin@example.com
Admin Password: secret123

It will take a while to run. Once it's done running, just run `rails s` to boot the server.

Let me know if you have any questions. Thanks!


jedieaston:
Have you checked out #1986? I did a similar thing, if you would like to take a peek.

octosteve:
@jedieaston I see that... I can't for the life of me figure out why I wrote this then!

PR #2096, opened 2022-09-19
Fix cookie samesite issue for LTI 1.3 deep linking

### LINKED ISSUE #1900: Cannot launch LTI 1.3. assignments in IFrame: Canvas fails to authenticate because cross-site cookie is blocked
### Summary:

In Google Chrome > 91, when using Canvas as an LTI 1.3 platform to launch External Tool assignments connected to an LTI 1.3. tool inside an IFrame, the launch fails with `error: login_required` and `error_description: Must have an active user session`.

It is noticeable that `log_session_id`, the cookie used by Canvas for session management, is blocked during the LTI 1.3. OIDC authentication flow on the request to the authorization endpoint `/api/lti/authorize_redirect`.  This behavior happens since [Chrome > 91 started enforcing the new SameSite policies](https://www.chromium.org/updates/same-site). That policy is blocking cross-site cookies without `SameSite=None` and `Secure` set, and redirects within IFrames [are considered cross-site](https://stackoverflow.com/questions/66671911/what-are-the-rules-for-samesite-lax-cookies-and-get-request-inside-iframe).

### Steps to reproduce:

 1. Have Google Chrome > 91 (or any other browser enforcing `SameSite=None` cookies for cross-site calls).
 1. Setup a development stack of Canvas 2021-05-26.01
 2. Configure [LTI 1.3.](https://canvas.instructure.com/doc/api/file.lti_dev_key_config.html)
 3. Configure an LTI tool (I am using a tool that is based on [`pylti1p3`](https://github.com/dmitry-viskov/pylti1.3)).
 4. Setup an external tool assignment.
 5. Uncheck the "launch in a separate tab" checkbox.
 6. Launch the assignment.
 7. Observe the LTI launch failing.  This can be done through the development console.

### Expected behavior:

The LTI launch should succeed.

### Actual behavior:

The LTI launch fails, here's a breakdown of the requests.

|#|From|To|Description|
|-|-|-|-|
|1. |*BROWSER*|CANVAS|Request assignment in Canvas, which is an External Tool (LTI).|
|2. |CANVAS|*BROWSER*|Reply with an empty IFrame and set the `log_session_id` cookie **without** `SameSite=None`. This means cross-site calls to CANVAS will not set the cookie.|
|3. |*BROWSER*|TOOL|Request `login/` from within the iframe. **This is cross-site.** It starts the OIDC flow. The frontend code will populate the iframe with the content of this request.|
|4. |TOOL|*BROWSER*|Reply with a redirect to CANVAS passing the LTI 1.3. auth parameters. This request **does not have `log_session_id` set**. It is blocked in Chrome (but Firefox allows it).|
|5. |*BROWSER*|CANVAS|Redirect to authenticate at Canvas is sent, it fails because we are not passing the `log_session_id` token.|
|6. |CANVAS|*BROWSER*|Reply with a redirect to TOOL, with an error.|
|7. |*BROWSER*|TOOL|This should be the final tool launch, but it is the error from the authentication.|
|8. |TOOL|*BROWSER*|Since it was an error, TOOL reply the "Invalid Tool Launch" message.|

Excerpt of the error redirect from `/api/lti/authorize`:

```
utf8: ✓
authenticity_token: <a base 64 encodied string that was omited by me>
error: login_required
error_description: Must have an active user session
state: state-891953f5-97dd-4891-837f-5d01958aeb29
```

### Additional notes:

Please, notice there are other resources in Forums and discussions around LTI 1.3. and cross-site cookies handling after the `SameSite=None` enforcing was added to browsers such as Chrome and Safari, but they are related to updating Tools that uses cross-site cookies.  This report is about to perform the LTI 1.3 authentication within an IFrame.

Examples:

1. https://community.canvaslms.com/t5/Canvas-Developers-Group/Safari-13-1-and-LTI-Integration/ba-p/273051
2. https://community.canvaslms.com/t5/Canvas-Developers-Group/SameSite-Cookies-and-Canvas/ba-p/257967

jbergfi:
LTI 1.3 deep linking don't seem to be working in latest Chrome. Seemingly this is due to because SameSite property of session and CSRF token cookies are not set to None to allow cross-site. Even though the deep link selection iframe ends up at same domain (Canvas) it's navigated to through the tool domain which effectively makes it cross-site.

Closes gh-1900

Test plan:
- Test LTI 1.3 deep linking with an external tool
- Test compatibility with different browsers (new browsers supporting SameSite: None and browsers not supporting)
- Investigate any undesired security implications this change might have

### DIFF
--- a/gems/canvas_breach_mitigation/lib/canvas_breach_mitigation/masking_secrets.rb
+++ b/gems/canvas_breach_mitigation/lib/canvas_breach_mitigation/masking_secrets.rb
@@ -32,7 +32,7 @@ module CanvasBreachMitigation
         encoded_masked_token = masked_token(unmasked_token(cookies["_csrf_token"]))
 
         cookie = { value: encoded_masked_token }
-        %i[domain httponly secure].each do |key|
+        %i[domain httponly secure same_site].each do |key|
           next unless options.key?(key)
 
           cookie[key] = options[key]
--- a/gems/request_context/lib/request_context/session.rb
+++ b/gems/request_context/lib/request_context/session.rb
@@ -32,7 +32,8 @@ module RequestContext
         ActionDispatch::Request.new(env).cookie_jar[:log_session_id] = {
           value: session_id,
           secure: Rails.application.config.session_options[:secure],
-          httponly: true
+          httponly: true,
+          same_site: Rails.application.config.session_options[:same_site]
         }
       end
 
--- a/lib/canvas/request_forgery_protection.rb
+++ b/lib/canvas/request_forgery_protection.rb
@@ -40,7 +40,7 @@ module Canvas
 
     def authenticity_token_options
       session_options = CanvasRails::Application.config.session_options
-      options = session_options.slice(:domain, :secure)
+      options = session_options.slice(:domain, :secure, :same_site)
       options[:httponly] = HostUrl.is_file_host?(request.host_with_port)
       options
     end


jbergfi:
Please consider I don't really know anything about Ruby or Canvas codebase or what impacts this change may have. This was merely my workaround how I could make my LTI 1.3 deep linking tool work with my self-hosted Canvas instance.

PR #2130, opened 2022-12-07
Fix broken check_for_docker_desktop in Japanese

CkReal:
The check_for_docker_desktop function did not work in the Japanese environment.
Therefore, The result of the LANG variable is used to branch the value of kMDItemKind.

How about this correction?


maths22:
This was replaced with `mdfind kind:application Docker.app`  Please let us know if that doesn't work in a Japanese locale.

PR #2136, opened 2023-01-03
Correct mangled try/catch block

Loccy:
This mangled try/catch block (no exception object specified to catch) causes the embedded chat to fail to load (as well as cause a number of issues for those of us who jump on the coat-tails of its iframe-resizing code in instutional learning content!). The simplest of fixes is proposed. Also see issue 2131.


maths22:
fixed in 4b37c2ce1bf4fb885db0e8e8646143a9d2d761d7

PR #2139, opened 2023-01-08
add non-null check

qq7:


eschiebel:
good catch. We will make this change in-house which will get reflected here once merged.
I will also take the opportunity to build the URL in a more robust way than string concatenation, just in cast `attachment.canvadoc_session_url` itself includes an origin.
Will reference the commit and close this PR when that's done.

jakeoeding:
@qq7 We just merged a fix yesterday for this: https://github.com/instructure/canvas-lms/commit/9bbb37354d1bc6644b6f93c25bd44048f1043297

PR #2146, opened 2023-01-26
React discussions: don't use optimisticResponse for initial post

jlebo:
Load entries over network after initial post when initial post is required to view replies

When using the updated react discussions interface, and a discussion requires an initial post before the user can view other replies, the replies often did not load after making the initial reply as the request to load replies was not waiting for the new reply to be created. This commit removes the optimisticResponse parameter to useMutation when creating a discussion entry while the user still requires an initial post to view replies. This causes the query for discussion entries to block until after useMutation returns so that the query will actually return all of the replies in the discussion. In other cases, optimisticResponse is used so that the UI is more responsive when adding discussion entries.

Test Plan:
With discussion redesign enabled...
1. Create a Discussion Topic with "reply before showing post" enabled
2. Make an initial post as the creator of the discussion.
3. View the discussion as a student - no replies should show
4. Make a post as the student - the discussion portion of the page should reload showing both the student and creator replies (two graphQL requests should be made on the network - one `CreateDiscussionEntry` and one `GetDiscussionQuery` - requests should be in that order and the second should not initiate until the first returns)
5. Make a second post as the student - the reply should show in the discussion without the entire discussion reloading (one graphQL request is made on the network: just `CreateDiscussionEntry`)


jlebo:
Made the fix in gerrit instead so it is easier to go through the code review process.

PR #2157, opened 2023-02-09
LtiDeepLinkingResponse can indicate that link open in new tab

matthewbennink:
Canvas defaults to using an iframe (new_tab: 0) for ltiResourceLink content items. This requires that a tool provider use the Canvas HTTP API to update each module item or assignment to specify that it open the LTI link in a new tab.

Since Canvas doesn't currently take the `window.targetName` into account when handling ltiResourceLink content items, it seemed OK to handle a value of "_blank" as a special case for indicating that the content should open in a new tab.

This would not affect any existing content and it's unlikely (?) to surprise any users who have set a targetName of "_blank". In fact, it may now do what they've intended it to do.

Test Plan:
  - Configure a course to launch an LTI Advantage Deep-Linking workflow
  - The tool should return a Deep-Linking response with at least one ltiResourceLink content item whose window.targetName value is set to "_blank"
  - Once the item populates (module item or assignment), ensure it opens in a new tab

### DIFF
--- a/app/controllers/lti/ims/concerns/deep_linking_modules.rb
+++ b/app/controllers/lti/ims/concerns/deep_linking_modules.rb
@@ -67,11 +67,19 @@ module Lti::IMS::Concerns
       }
     end
 
+    # the window property in a deep linking response can contain
+    # link-specific options. If a targetName is present, it can
+    # be used to determine whether the link should default to
+    # being opened in a new tab or not
+    def open_in_new_tab?(content_item)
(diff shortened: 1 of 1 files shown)


matthewbennink:
Relevant part of the LTI Deep-Linking spec: https://www.imsglobal.org/spec/lti-dl/v2p0#lti-resource-link. See the `window` property in particular.

xandroxygen:
@matthewbennink thanks for your contribution! This change has been merged and will be deployed to production on April 12th

PR #2172, opened 2023-03-17
Observer enrollments need to be touched with latest sis batch id

cgravatt:
Observer enrollments not specificied in an enrollments sis import run in batch mode are not updated with the latest sis batch id. As a result, the enrollments are deleted even if the student enrollment remains active

Test plan
  - Set up an observer with an active enrollment tied to a student
  - Run an enrollment sis import in batch mode supplying the student enrollment as active
  >> Expect the observer enrollment to still be active

### DIFF
--- a/lib/sis/enrollment_importer.rb
+++ b/lib/sis/enrollment_importer.rb
@@ -341,6 +341,13 @@ module SIS
             @roll_back_data << data if data
           else
             @enrollments_to_update_sis_batch_ids << enrollment.id
+            # Make sure linked observer enrollments are added to be touched with sis_batch_id
+            linked_enrl = Enrollment.where.not(sis_batch_id: nil)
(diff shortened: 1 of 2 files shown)


dustin-cowles:
@cgravatt This was addressed using a different method in https://github.com/instructure/canvas-lms/commit/dad3f86eef72a56d8e6af3915304b88dc2ef625e

cgravatt:
@dustin-cowles are you all reviewing and evaluating #2170 as well? I only mention it here because for us these two issues are related. The bug addressed in dad3f86 caused a lot of observer enrollments to get marked inactive that subsequent runs of the user observer sis import process cannot address without intervention.

jstanley0:
@cgravatt I'm working on that one now. Currently running your commit through the test suite

PR #2188, opened 2023-04-14
Update updated_at on sis batch when progress is updated

cgravatt:
When the progress is updated on an sis batch via fast_updated_progress the updated_at field should also be updated.

The updated_at field is useful when monitoring when an import might be hung. For example, progress hasn't incremented since x time. Currently, as far as i've observed the updated_at value never progresses during the cleanup batch phase.

Test plan
  - Run an enrollment batch import and validate the updated_at field is changing as the progress changes.

### DIFF
--- a/app/models/sis_batch.rb
+++ b/app/models/sis_batch.rb
@@ -278,7 +278,7 @@ class SisBatch < ActiveRecord::Base
 
     self.progress = val
     state = SisBatch.connection.select_value(<<~SQL.squish)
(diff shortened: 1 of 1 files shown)


jstanley0:
@cgravatt thanks for the submission. I had to tweak it slightly since the interpolation into sql caught the attention of our sql injection scanner, even though the interpolation was already there and none of the values are under user control anyway

cgravatt:
Makes sense. Thanks @jstanley0

PR #2189, opened 2023-04-17
fix temp uploaded files path

amg-web:
fix temp uploaded files path
By default it uploads files to /tmp/attachment_fu

in case of several canvas instances issue with file upload due to permissions issue
changing folder to RWX for all users does not fix issue

we change to tmp/attachment_fu 
in this case tmp uploaded files are in project's folder tmp and no issues

### DIFF
--- a/gems/attachment_fu/lib/attachment_fu/railtie.rb
+++ b/gems/attachment_fu/lib/attachment_fu/railtie.rb
@@ -25,7 +25,7 @@ module AttachmentFu
     end
 
     def self.setup_tempfile_path
-      AttachmentFu.tempfile_path = "/tmp/attachment_fu"
+      AttachmentFu.tempfile_path = "tmp/attachment_fu"
       AttachmentFu.tempfile_path = ATTACHMENT_FU_TEMPFILE_PATH if Object.const_defined?(:ATTACHMENT_FU_TEMPFILE_PATH)
 
       begin


gems/attachment_fu/lib/attachment_fu/railtie.rb
```diff
@@ -25,7 +25,7 @@ class Railtie < ::Rails::Railtie
     end
 
     def self.setup_tempfile_path
-      AttachmentFu.tempfile_path = "/tmp/attachment_fu"
+      AttachmentFu.tempfile_path = "tmp/attachment_fu"
```
maths22, line 28: We would prefer not to change the default here, but making this configurable would seem reasonable.  Currently it looks like you can set `ATTACHMENT_FU_TEMPFILE_PATH` in initialization code; maybe making it also support an environment variable or reading it from a YML file would be reasonable.

maths22 (changes requested):

gems/attachment_fu/lib/attachment_fu/railtie.rb
amg-web: project uses dir: canvasdir/tmp for all temporary and uploaded files (if no s3), but uses system temp for uploaded temp files.

Sorry, I'm admin, not ruby developer. So I can not do more seriuous code changes.
Otherwise I whould offer redis gem update. But when I try newer version, some more code needs to be changed.

PR #2203, opened 2023-05-09
Update password_policy.rb

gdenne:
Updated Common Password List to use Top 100 weak passwords from Wikipedia. Previous list was from 2012 and outdated.


gdenne:
Added CC BY-SA link and removed profanity censored passwords from list @dustin-cowles

PR #2209, opened 2023-05-17
mr-n30 idor fix media tracks controller

mr-n30:
Please see Bugcrowd for PoC of this vulnerability.

- Best regards.


rmsy:
Hi @mr-n30, security vulnerabilities should only be reported through Bugcrowd -- never publicly. If possible, could you please delete your fork with the fix until we've had time to review and address your report?

PR #2223, opened 2023-06-27
Update password policy

gdenne:
Updated common password policy to only include passwords >8 characters due to NIST policy - Canvas does not allow passwords under 8 characters so much of the list was redundant because it had passwords lower than 8. I also changed the password example in the UI to a commonly known bad example password because the UI error message was a bit confusing in showing 'password' when talking about password policy.


dustin-cowles:
Hi @gdenne, thanks for your contributions! This commit is based on a very old and non-merged commit. Can you please rebase it on the tip of master?
Thanks!
Dustin

gdenne:
Hi @dustin-cowles Sorry, my bad. Let me give that another go.

gdenne:
@dustin-cowles PR updated, thank you for your help!

dustin-cowles:
It is possible for a Canvas instance to be configured to not allow common passwords and to allow passwords <8 characters.

PR #2224, opened 2023-06-28
Update content.rb

sparkyrider:
Quick typo fix gaurentee -> guarantee


sparkyrider:
Noticed a Quick typo in "lib/api/html/content.rb". No other changes, simple update.

PR #2225, opened 2023-06-28
Update bad password list/policy

gdenne:
Updated common password policy to only include passwords > 8 characters due to NIST policy - Canvas does not allow passwords under 8 characters so much of the list was redundant because it had passwords lower than 8. I also changed the password example in the UI to a commonly known bad example password because the UI error message was a bit confusing in showing 'password' when talking about password policy.


dustin-cowles:
duplicate of #2223

Issue #2199, opened 2023-05-06
Permission issues for dev_docker_setup.sh on Linux Ubuntu and dory setup FIX


mr-n30:
### Summary:
When running the script `./script/docker_dev_setup.sh`  on Ubuntu 20.04 with 200 GB for storage and 16GB RAM the error below is returned:
```
> react-test-renderer@17.0.2" has incorrect peer dependency "react@17.0.2".
warning " > @storybook/react@6.5.9" has unmet peer dependency "require-from-string@^2.0.2".
warning " > @instructure/babel-plugin-themeable-styles@7.19.0-canvas" has incorrect peer dependency "webpack@^3 || ^4".
warning "workspace-aggregator-4228077c-1de3-4c6b-b32d-7d1def137de3 > @instructure/canvas-rce > axe-testcafe@3.0.0" has incorrect peer dependency "axe-core@>=2.2.3 <4".
error Could not write file "/usr/src/app/yarn-error.log": "EACCES: permission denied, open '/usr/src/app/yarn-error.log'"
error An unexpected error occurred: "EACCES: permission denied, mkdir '/usr/src/app/packages/babel-plugin-themeable-styles/node_modules'".
info Visit https://yarnpkg.com/en/docs/cli/install for documentation about this command.
  /o\ Something went wrong. Check /home/ubuntu/canvas-lms/log/docker_dev_setup.log for details.
```

This issue is with yarn not being able to write to certain directories and files. During the installation process before `yarn install` is called the owner and permissions are changed which causes this error. I've added a fix for Ubuntu 20.04 host operating systems in the Additional notes section of this issue.

### Steps to reproduce:

 1. Install Ubuntu Server 20.04 on a VM with 200GB storage and 16GB RAM
 2. Clone canvas-lms repo from GitHub and run: `./script/docker_dev_setup.sh`
 3. See that you get the error message from above

### Expected behavior:
No errors should occur

### Actual behavior:
Permission errors via `yarn install`

### Additional notes:
Below are the steps I took to fix the issues:

Replace the lines:

```diff
diff --git a/Dockerfile b/Dockerfile
index bda70372d8..a61e8f3ef5 100644
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,6 +29,16 @@ WORKDIR $APP_HOME
 
 USER root
 
+# Add docker to group sudo so we can use
+# `sudo chown` later in yarn install and
+# prevent permission issues.
+# Create a new user with sudo privileges
[...]

PR #2227, opened 2023-06-29
docker_dev_setup.sh permission fix for issues in https://github.com/instructure/canvas-lms/issues/2199

mr-n30:
This pull request fixes the issues in: 
- `https://github.com/instructure/canvas-lms/issues/2199`

When trying to install Canvas via `docker_dev_setup.sh` the user is unable to complete the installation script due to permissions of files and directories in the application root being changed resulting in other programs being unable to install packages and dependencies in those locations.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -29,6 +29,12 @@ WORKDIR $APP_HOME
 
 USER root
 
+# This is required in order to change the permissions and 
+# ownership of the directory that causes permission issues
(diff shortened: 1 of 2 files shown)


maths22:
I don't think we want to solve this issue in the way proposed here.

Have you tried following these directions for running on linux? https://github.com/instructure/canvas-lms/tree/master/doc/docker#development

xeruf:
tried all this, still experiencing that error two years later

PR #2335, opened 2024-03-25
OIDC login hint

chris-perts:
This adds some support for the `login_hint` parameter, defined by the [OIDC Core 1.0 spec](https://openid.net/specs/openid-connect-core-1_0.html#AuthRequest) and mentioned in [Google's documentation](https://developers.google.com/identity/openid-connect/openid-connect#login-hint) of the same.

With this change, if users are directed to "/login/google?login_hint=user@foo.com", assuming canvas has the "google" authentication option set up, it will add `&login_hint` to the authorization request URL. The effect on the user is to skip the account choice screen if the user has multiple active sessions with the IdP and the email provided in `login_hint` matches one of them.


CodeCraftPro1:
Bro what? No shot this works. DELETE THIS 🙄🙄🙄

ccutrer:
This has already been implemented as of 80425ec494d64f0db86677a438932f451c231c48

PR #2348, opened 2024-04-23
fix: correct invalid selector in revert "what-if" scores handler

Cooldogyum:
You were unable to revert "What-If" scores in the gradebook due to an invalid query selector.

Test Plan:
   - Verify the "Revert to Actual Score" button is functional after setting "What-If" scores


dustin-cowles:
@Cooldogyum thanks for your contribution, however this has already been merged as 07bd4d61d39a8f165d81b1161ee768bc557d4e5c

PR #2356, opened 2024-05-29
Fix margin issue on access code input

ashermyers:
Before:
image

After:
image

Fixes margin issue with input not properly aligning with button on the access code input page.


slaughter550:
@ashermyers ^ was this intentional? We’re currently checking with product to see if this change is desirable

ashermyers:
New PR is here https://github.com/instructure/canvas-lms/pull/2358 (Closed this one by accident) 

It seems to not be the intended behavior. All inputs in the design theme have extra margin added to space between other form elements. That's why I submitted the PR is to fix this since this form uses a side-by-side and not a stacked layout like generic form elements.

PR #2358, opened 2024-05-31
Fix margin issue on access code input

ashermyers:
Before:
image

After:
image

Fixes margin issue with input not properly aligning with button on the access code input page.


ashermyers:
It seems to not be the intended behavior. All inputs in the design theme have extra margin added to space between other form elements. That's why I submitted the PR is to fix this since this form uses a side-by-side and not a stacked layout like generic form elements.

spencerolson:
Thanks for your contribution, @ashermyers ! This has been merged in https://github.com/instructure/canvas-lms/commit/ca74d1c56f0ddd48975092c0e7e035ba77ca428d .

PR #2459, opened 2025-02-20
Make 'new_window' a valid display type

julianvu:
LTI tools registered using Dynamic Registration have no way of configuring navigation placements to open in a new tab instead of in frame. Adding 'new_window' to enum of VALID_DISPLAY_TYPES in InternalLtiConfiguration schema allows for this.

Test Plan:
- Configure an LTI 1.3 tool with the Course Navigation placement using Dynamic Registration
- Enable the tool in a course
- Select the course navigation link added by the tool. The tool should open in a new tab instead of in frame.

### DIFF
--- a/lib/schemas/internal_lti_configuration.rb
+++ b/lib/schemas/internal_lti_configuration.rb
@@ -22,7 +22,7 @@ module Schemas
   # Represents the "internal" JSON schema used to configure an LTI 1.3 tool,
   # as stored in Lti::ToolConfiguration and used in Lti::Registration.
   class InternalLtiConfiguration < Base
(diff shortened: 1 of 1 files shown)


julianvu:
The `new_window` display type is already supported internally to open LTI 1.3 tools in a new tab. The issue is `new_window` triggers a validation error if it's included in the tool registration request. Adding it to the enum will allow this parameter through, which will then allow a tool registered using Dynamic Registration to launch in a new tab.

spencerolson:
Thanks for the pull request, @julianvu ! I'm running this by the owning team for code review now. I'll keep the labels updated with the current status🫡

julianvu:
@xandroxygen Thanks for picking up the fix and getting it merged in! It is very much appreciated!

PR #2473, opened 2025-03-18
sc-9514 Add Ruby Inline PDF to Canvas

therealjasonkenney:
This adds the Ruby Inline PDF to Canvas.

* I put this in a new file: `Gemfile.d/learn.rb` so it isolates any gems we put in that are specific to learn.
* This is not production ready - I've made this PR for testing purposes, and to make it easier to pull in my branch and test this locally and on staging. Before it goes to production, the gem PR needs to be merged, tagged (and bagged), published and then this PR needs to be modified to use the versioned gem instead of the github sha.

## Testing
I've currently been using Firefox 136 on an M1 Macbook Pro running MacOS 15.

* Run `bundle install`
* Ensure you have a PDF within a course or module.
* Verify Canvas behaves as before checking Files and Modules.
* In the plugins section, you should see `Inline PDF` enable it.
* Verify Files within Canvas behaves as before, but Modules should now render a pdf, you may need to fiddle with browser cache, it might load the prior rendered view. Make sure to also view as Student.
* Go back to the Inline PDF Plugin Configuration and check the 'Override File Preview' setting.
* You may need to fiddle with cache or logout/login. 
* Verify Files now renders pdfs inline.

**NOTE:** Nothing has been tested regarding Assignments at the moment. (Details on that behavior will be added soon)


amg-web:
interesting
but can not open the respective plugin setting - error 500 for page /plugins/canvas_inline_pdf
canvas **origin/stable/2025-05-07**

dustin-cowles:
Closing due to lack of activity, feel free to update and reopen.

PR #2532, opened 2025-08-17
Allow time tag to retain 'datetime' attribute in RCE

seanbeirnes:
# Overview

The CanvasSanitize configuration previously allowed the `<time>` tag but stripped its `datetime` attribute. This caused semantic information about dates/times to be lost when user content was sanitized. The proposed change updates CanvasSanitize to preserve the `datetime` attribute on the `<time>` tag. Corresponding RSpec coverage was added to ensure the attribute is retained exclusively on `<time>`. The TinyMCE config for the Rich Content Editor (RCE) was also updated to reflect this added support on the client side.

### Preserving Semantics

The `datetime` attribute is an important part of the `<time>` element in HTML, allowing dates/times to be machine-readable by assistive technologies, user agents, and scripts. Stripping the attribute undermines the element’s intended semantic use.

Source: [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/time)

### Accessibility

This change improves conformance with [WCAG 2.2 SC 1.3.1 Info and Relationships (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/info-and-relationships.html) by preserving the semantic relationship between the visible date text and its machine-readable representation. This helps screen readers and other tools interpret temporal information more accurately.

For example, the following HTML snippet semantically uses the `<time>` tag and `datetime` attribute to remove ambiguity for visually impaired users. It is a *duration* of 12 minutes and 30 seconds, not 12 hours and 30 minutes or a time of day.
```html
<p>Lesson 1 - An Introduction (<time datetime="PT12M30S">12:30</time>)</p>
```

The ISO 8601 datetime string above maps to the following:
P = "period"
T = "time component"
12M = "12 minutes"
30S = "30 seconds"

## Proposed Change

* Updated **CanvasSanitize** configuration to allow the `datetime` attribute **only** on `<time>` elements.
* Updated **defaultTinymceConfig.ts** so the RCE also supports `datetime` on `<time>`.

## Test Plan

* Added new spec to **canvas_sanitize_spec.rb** to verify that `<time datetime="...">` retains its attribute while other elements (`<div datetime>`, `<span datetime>`) have it stripped.


seanbeirnes:
I amended the commit to update the expected TinyMCE config test in **defaultTinymceConfig.test.js** so it would pass.

One test is still failing due to **QuotasTabContent.tsx**. This file is unrelated to my changed files.

PR #2545, opened 2025-10-02
Fix docker_dev_setup script on Linux: status check

kevinroleke:
On systemd machines without the `service` command available, docker will always appear to not be running.

The file `script/common/os/linux/impl.sh` will attempt to run the following command to check the status of docker:
```bash
sudo systemctl docker status # this returns an error, always--the arguments are in the wrong order
```

I suggest we use `docker info` instead. This command will return an error code if docker is inactive and success if docker is active. Alternatively, the arguments should be properly arranged for `service` vs `systemctl`.

### DIFF
--- a/script/common/os/linux/impl.sh
+++ b/script/common/os/linux/impl.sh
@@ -15,7 +15,7 @@ function set_service_util {
 }
 
 function start_docker_daemon {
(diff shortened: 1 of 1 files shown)


oshanz:
openSUSE Tumbleweed

image

PR #2580, opened 2026-01-29
fix: reduce side effect logic

duidae:
1. filter() should be a pure function, adding extra logic inside filter() could lead to unexpected behavior/bug, so did a small refinement to it
2. minor codebase refinement 
thanks!


dustin-cowles:
Looks good to me, if you can get the conflicts resolved I can submit it to the owning team for review.

duidae:
@dustin-cowles conflicts are fixed, please have a look & thanks! 🙂

PR #2588, opened 2026-02-23
Preserve integration info during course clone

elenabaurkot:
Add opt-in settings to copy integration_id and integration_data on assignments during course copy. External systems rely on these fields to correlate Canvas assignments with third-party records and assignments across courses can share the same IDs.

We reuse intergration_ids and integration_data across courses to be able to track assignment data between courses and semesters. I opened a [support ticket](https://adminconsole.canvaslms.com/s/case/500TU00000rMoR3YAK/copy-integration-ids-set-on-assignments-from-one-course-to-another-on-course-clone) and was told integration_data should copy over and that integration_ids were meant to be unique, but since these link to third parties was hopeful people could opt into maintaining over course clone in case the same courses need to link to the same third party data for our case. 

Our data team uses DAP (data access platform) to access Canvas data and it only contains integration_id for assignments not integration_data. If we are able to clone only integration_data on course clone, could this potentially be added to DAP? 

Test Plan:
* Create a course with assignments that have integration_id and integration_data set
* Copy the course with both options enabled
* Verify cloned assignments retain the values
* Copy again with options disabled
* Verify fields are cleared (default behavior)
* Test via API: POST course copy with settings[copy_integration_ids]=true
* Verify same behavior as UI

### DIFF
--- a/spec/models/content_migration/course_copy_assignments_spec.rb
+++ b/spec/models/content_migration/course_copy_assignments_spec.rb
@@ -1199,5 +1199,50 @@ describe ContentMigration do
         expect(a_to).to be_valid
       end
     end
+
+    context "integration info copying" do
+      it "copies integration_id and integration_data when copy_integration_info is enabled" do
+        assignment_model(course: @copy_from, points_possible: 10, submission_types: "online_text_entry")
+        @assignment.update!(integration_id: "ext-system-42", integration_data: { "sis_source" => "ABC", "vendor_id" => 99 })
+
+        @cm.migration_settings[:copy_integration_info] = true
+        @cm.save!
+
+        run_course_copy
+
+        a_to = @copy_to.assignments.find_by(migration_id: mig_id(@assignment))
+        expect(a_to.integration_id).to eq "ext-system-42"
+        expect(a_to.integration_data).to eq({ "sis_source" => "ABC", "vendor_id" => 99 })
+      end
+
+      it "does not copy integration_id or integration_data when copy_integration_info is not enabled" do
+        assignment_model(course: @copy_from, points_possible: 10, submission_types: "online_text_entry")
+        @assignment.update!(integration_id: "ext-system-42", integration_data: { "sis_source" => "ABC", "vendor_id" => 99 })
+
+        run_course_copy
+
+        a_to = @copy_to.assignments.find_by(migration_id: mig_id(@assignment))
+        expect(a_to.integration_id).to be_nil
+        expect(a_to.integration_data).to be_nil
+      end
+
+      it "copies only populated fields when copy_integration_info is enabled" do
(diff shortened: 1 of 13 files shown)


viszpis:
👋 One more thing, I think this feature needs to be behind a Feature Flag. The flag should be defined in config/feature_flags/00_standard.yml Something like this

```
course_copy_allow_copying_integration_info:
  applies_to: SiteAdmin
  state: hidden
  display_name: Course Copy: allow copying integration_id and integration_data
  description: Adds opt-in settings to copy integration_id and integration_data on assignments during course copy
  environments:
    ci:
      state: allowed_on
    development:
      state: allowed_on
```

This should control the import/export flow plus the UI rendering (the checkbox rendering) part.

elenabaurkot:
Thanks for the feedback @viszpis, I've updated the PR with the combined setting, feature flag and environments block. Quick question: once this merges, what's the typical process for enabling a SiteAdmin flag in production? We'd love to be able to use this feature on our instance.

spec/models/content_migration/course_copy_assignments_spec.rb
```diff
@@ -1199,5 +1199,50 @@
...
+
+      it "does not copy integration_id or integration_data when copy_integration_info is not enabled" do
+        assignment_model(course: @copy_from, points_possible: 10, submission_types: "online_text_entry")
+        @assignment.update!(integration_id: "ext-system-42", integration_data: { "sis_source" => "ABC", "vendor_id" => 99 })
+
+        run_course_copy
+
+        a_to = @copy_to.assignments.find_by(migration_id: mig_id(@assignment))
+        expect(a_to.integration_id).to be_nil
+        expect(a_to.integration_data).to be_nil
```
spencerolson, line 1226: this assertion is failing with:

```
Failure/Error: expect(a_to.integration_data).to be_nil

 expected: nil
      got: {}
```

spec/models/content_migration/course_copy_assignments_spec.rb
```diff
@@ -1199,5 +1199,50 @@
...
+        a2 = @copy_from.assignments.create!(title: "Only Data", submission_types: "online_text_entry", integration_data: { "data" => "only" })
+
+        @cm.migration_settings[:copy_integration_info] = true
+        @cm.save!
+
+        run_course_copy
+
+        a1_to = @copy_to.assignments.find_by(migration_id: mig_id(a1))
+        expect(a1_to.integration_id).to eq "id-only"
+        expect(a1_to.integration_data).to be_nil
```
spencerolson, line 1240: this assertion is failing with:

```
Failure/Error: expect(a1_to.integration_data).to be_nil

 expected: nil
      got: {}
```

spencerolson (changes requested):
Hi @elenabaurkot this generally looks good, but currently the test suite is failing. Two of the tests you added have failing assertions. Can you fix those please? Thanks!

PR #2611, opened 2026-03-17
Discussion Replies - Edit/Delete Triggers Live Event

braysch:
Editing or deleting a reply as part of a discussion thread will trigger a live event.
Includes automated tests.
Tested manually with Kinesis server.

Created new discussion_entry_updated function as one did not exist.

### DIFF
--- a/lib/canvas/live_events_callbacks.rb
+++ b/lib/canvas/live_events_callbacks.rb
@@ -116,7 +116,11 @@ module Canvas::LiveEventsCallbacks
       end
       Canvas::LiveEvents.course_updated(obj)
     when DiscussionEntry
-      Canvas::LiveEvents.discussion_entry_updated(obj)
+      if changes["deleted_at"] && obj.deleted?
+        Canvas::LiveEvents.discussion_entry_deleted(obj)
+      else
+        Canvas::LiveEvents.discussion_entry_updated(obj)
+      end
     when DiscussionTopic
       Canvas::LiveEvents.discussion_topic_updated(obj)
     when Enrollment
--- a/lib/canvas/live_events_callbacks.rb
+++ b/lib/canvas/live_events_callbacks.rb
@@ -115,6 +115,8 @@ module Canvas::LiveEventsCallbacks
         Canvas::LiveEvents.course_syllabus_updated(obj, changes["syllabus_body"].first)
       end
       Canvas::LiveEvents.course_updated(obj)
+    when DiscussionEntry
+      Canvas::LiveEvents.discussion_entry_updated(obj)
     when DiscussionTopic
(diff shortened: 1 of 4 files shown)


spencerolson:
Hi @braysch , the event is exactly the same when a entry is edited or deleted, so there's no way to differentiate between a delete or edit event. In both cases, the event name is `"discussion_entry_updated"` and the event payload looks like:

```
{
  user_id: 23,
  created_at: 2026-04-14 11:08:14.440959000 MDT -06:00,
  discussion_entry_id: 51,
  discussion_topic_id: 106,
  text: "<p>some reply<span class=\"post_mention\"> here</span></p>"
}
```

Can you please make the event name `"discussion_entry_deleted"` for deletion events? That will allow live event consumers to differentiate between the two. Note that when a discussion entry is deleted in the UI, it's actually soft-deletion that's happening, which under-the-hood is an update action, not a destroy action. So you would just modify your existing code in the after_update method to look something like:

```
    when DiscussionEntry
      if changes["deleted_at"] && obj.deleted?
        Canvas::LiveEvents.discussion_entry_deleted(obj)
      else
        Canvas::LiveEvents.discussion_entry_updated(obj)
      end
```

and then you'd create a `Canvas::LiveEvents.discussion_entry_deleted` method which sets the event name to `"discussion_entry_deleted"`.

Let me know if you have any questions, and thanks!

lib/canvas/live_events_callbacks.rb
```diff
@@ -115,6 +115,8 @@ def self.after_update(obj, changes)
         Canvas::LiveEvents.course_syllabus_updated(obj, changes["syllabus_body"].first)
       end
       Canvas::LiveEvents.course_updated(obj)
+    when DiscussionEntry
+      Canvas::LiveEvents.discussion_entry_updated(obj)
```
spencerolson: Can you please make the event name `"discussion_entry_deleted"` for deletion events? That will allow live event consumers to differentiate between the two. Note that when a discussion entry is deleted in the UI, it's actually soft-deletion that's happening, which under-the-hood is an update action, not a destroy action. So you would just modify your existing code in the after_update method to look something like:

```
    when DiscussionEntry
      if changes["deleted_at"] && obj.deleted?
        Canvas::LiveEvents.discussion_entry_deleted(obj)
      else
        Canvas::LiveEvents.discussion_entry_updated(obj)
      end
```

and then you'd create a `Canvas::LiveEvents.discussion_entry_deleted` method which sets the event name to `"discussion_entry_deleted"`.

spencerolson (changes requested):

lib/canvas/live_events_callbacks.rb
braysch: Added a separate `discussion_entry_deleted` event.

PR #2612, opened 2026-03-17
Live events for "classic quiz with no assignment" content type + live events for all quiz types content update + new event types for quizzes

JoshBothell:
- Live events for "classic quiz with no assignment" content type
- Live events for all quiz types content update
- New event types for quizzes to avoid confusion regarding using "assignment_updated" events for quizzes with no assignment. 

New live event types are `quiz_created` and `quiz_updated`, they were added to avoid confusion when sending live events for quizzes that have no associated assignment. They will only fire in these edge cases, for standard classic quizzes, the live event will still be of type `assignment_[ updated | created ]`.


JoshBothell:
** EDIT: we sorted it out on our end, you can disregard this. CLA assistant is happy now** 

Hi @maths22,
We are running into an issue with the CLA assistant. My colleague Bradan Schwanke authored some commits early in this branch, before he had his local git configured to have an email. He has signed the CLA on his normal github, but we can't seem to find a way to sign it as "Bradan Schwanke," as that author has no associated email. Please let me know if this is something you need us to sort through (ammending old commits or something), or if this is just something you can override given that Bradan has signed the CLA. To be clear, the earlier commits were from Bradan (@braysch), the assistant just doesn't recognize the commits that were authored with no email. 

Best,
Josh

JoshBothell:
Hi @maths22 

I am sure things are pretty hectic at Instructure right now. However, now that this PR has CLA and merge conflicts taken care of, can you give me a ballpark estimate on when we can expect an update on these changes being merged in?

Best,
Josh

jasonfill:
@JoshBothell After making [our changes](https://github.com/instructure/canvas-lms/pull/2646) I realized this PR was already out there and very similar. The current quizzes live event implementation doesn't send an event if any quiz-specific fields are updated (password, etc.), which was the primary driver for our change.

However, I see you're also looking for a `quiz_updated` event when a question is added or updated. Curious what your thoughts are here. Would it be better to have `quiz_question_added` and `quiz_question_updated` as separate events, to differentiate core quiz data updates from question-level changes? For our use case that would be preferable, since we don't need to be notified of question-level changes. Other implementations (potentially including yours) may need both levels of granularity though.

Since these are new events, it'd be great if the Quizzes team could weigh in, so we can align on an implementation that ideally carries forward into New Quizzes as well.

JoshBothell:
@jasonfill When we started work on increasing live event coverage, we were told by Instructure that when possible, use existing event types rather than creating new ones. That said, I fully agree with your comment and am happy to make the changes. I think the most important thing here is parity with the behavior of New Quizzes. I'll let them make the call on how this is implemented and I will make the necessary changes to my PR to align with the behavior of New Quizzes. 

Thank you for your input! We appreciate seeing progress on the live events work and I am thrilled to hear that New Quizzes may get the same live event we are trying to add in this PR.

Issue #2512, opened 2025-06-12
Dark Mode feature


Dani-DEV28:
Looking to add a dark mode in the webversion of Canvas

sharktide:
Please do

PR #2617, opened 2026-03-24
Add user dark mode toggle

bell-kevin:
### Motivation
- Provide a simple, user-facing dark mode option for the app shell so non-InstUI parts are readable in dark palettes.
- Persist the user preference across page loads so toggling dark mode is a one-time action for users.

### Description
- Updated `ui/boot/index.js` to add a floating dark mode toggle button that appears after `DOMContentLoaded` and toggles between `Enable dark mode` / `Disable dark mode`.
- Added localStorage persistence using the key `canvas_theme_preference` and restore logic on startup to reapply the chosen theme.
- Added `setDarkModeStyle`, `applyDarkMode`, and `renderDarkModeToggle` to inject lightweight global dark CSS and to drive InstUI career-dark theme overrides via `loadCareerTheme` when dark mode is enabled.
- Scoped changes to only runtime JS so existing stylesheets are not modified; the implementation falls back safely when `career` theme resources are not available.


bell-kevin:
https://github.com/instructure/canvas-lms/issues/2512

PR #2620, opened 2026-03-30
Assignments copied between courses now default to unpublished

hergin:
When copying an assignment from one course to another, the copied assignment was automatically published in the new course. This change makes copied assignments default to an unpublished state instead, giving instructors a chance to review the assignment before it becomes available to students. If an assignment's workflow state is explicitly set in the source course, that state is preserved during the copy.

Test Plan:
  - Create a published assignment in a source course
  - Copy the course to a new course
  - Verify the copied assignment is unpublished in the new course
  - Verify that if an unpublished assignment is copied, it remains unpublished


hergin:
Why is this still pending? @spencerolson

PR #2628, opened 2026-05-01
School Project Pull Request

Oliphant714:
What exactly does this commit do?


Oliphant714:
I'm curious about learning model systems and I'm new to most of the languages used here.  I agree with the changes!

simonista:
@0xfbad please keep the conversation here positive and constructive.

PR #2640, opened 2026-06-06
Add GET/LIST API for Quiz Extensions and NQ Accommodations

jasonfill:
The Quiz Extensions and New Quizzes Accommodations APIs currently only support creating extensions (POST), but provide no way to read them back. This makes it impossible for integrations to query what extensions have been granted to students, and there is no event-driven notification when extensions are set through the UI.
 
This change adds:
 
1. GET /api/v1/courses/:course_id/quizzes/:quiz_id/extensions
   - Lists quiz extensions from quiz_submissions where extra_attempts, extra_time, or manually_unlocked are set
   - Supports ?user_id= filter and pagination
   - Requires manage_grades or manage_assignments permission
 
2. GET /api/v1/courses/:course_id/new_quizzes/:assignment_id/accommodations
   GET /api/v1/courses/:course_id/new_quizzes/:assignment_id/accommodations/:user_id
   - Proxies to the New Quizzes service at /api/assignments/:id/participants
   - Authenticates via ServicesJwt, uses Canvas.timeout_protection
   - Returns extra_time, extra_attempts, reduce_choices_enabled
   - Returns 502 for upstream failures (timeout, invalid JSON, non-200)
 
3. Live Events
   - quiz_extension_created: emitted from QuizExtensionsController and CourseQuizExtensionsController after extensions are saved for Classic Quizzes
   - Uses global IDs (Switchman) for cross-shard uniqueness
  
> [!NOTE]
> This does NOT create live events for NQ as that is a closed system and I could not see a way to do that properly.  
 
flag=none
 
Test Plan:
- As an admin, create a classic quiz with students enrolled
- POST to create quiz extensions for a student (extra_attempts: 2)
- GET /api/v1/courses/:id/quizzes/:id/extensions
  - Verify response contains the extension with integer user_id/quiz_id
  - Verify students without extensions are excluded
  - Test ?user_id= filter returns only that student
- As a student, GET the extensions endpoint
  - Verify 403 Forbidden is returned
- Create a New Quizzes (quiz_lti) assignment
- POST to set accommodations (extra_time, extra_attempts)
- GET /api/v1/courses/:id/new_quizzes/:assignment_id/accommodations
  - Verify accommodations are returned from the NQ service
- GET /api/v1/courses/:id/new_quizzes/:assignment_id/accommodations/:user_id
  - Verify single user accommodation is returned
  - Verify 404 for a user with no accommodation
- Verify live events are emitted in the live events stream when extensions are created via the API
- Run RSpec:
  bin/rspec spec/apis/v1/quizzes/quiz_extensions_index_api_spec.rb \
    spec/apis/v1/quizzes/new_quiz_accommodations_api_spec.rb \
    spec/lib/canvas/live_events_quiz_extensions_spec.rb \
    spec/apis/v1/quizzes/quiz_extensions_api_spec.rb

### DIFF
--- a/config/locales/ko.yml
+++ b/config/locales/ko.yml
@@ -33,7 +33,7 @@ ko:
     title: "승인되지 않음"
   4th_grade_fdfaa9a9: 4학년
   558x270_png_jpg_gif_c3faa839: "558x270 png, jpg, gif"
-  558x558_png_jpg_gif_1_8x_the_standard_tile_size_so_c56cc472: "558x558 png, jpg, gif (표준 타일 크기의 1.8배이며, 필요에 따라 확대 혹은 축소될 수 있습니다.)"
+  558x558_png_jpg_gif_1_8x_the_standard_tile_size_so_c56cc472: "558x558 png, jpg, gif (표준 타일 크기의 1.8배이므로 필요에 따라 확대 혹은 축소할 수 있습니다.)"
   5th_grade_ea81bdea: 5학년
   6th_grade_d30c812f: 6학년
   7th_grade_c477956c: 7학년
@@ -49,11 +49,11 @@ ko:
     drop_lowest_rule: 
       one: "%{count}개 최저 점수 제거"
       other: "%{count}개 최저 점수 제거"
-    never_drop_rule: "%{assignment_name}을(를) 드롭한 적 없음"
+    never_drop_rule: "%{assignment_name} 제거하지 않음"
     rules_text: 
       one: "%{count}개 규칙"
       other: "%{count}개 규칙"
-    toggle_message: "과제 표시 토글"
+    toggle_message: "과제 표시 여부 토글"
   AssignmentKeyBindingsMixin: 
     keyboard_add_item: "선택한 그룹에 과제 추가"
     keyboard_close_menu: "활성 대화 닫기"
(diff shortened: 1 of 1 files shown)


app/controllers/quizzes/new_quiz_accommodations_controller.rb
```diff
@@ -0,0 +1,229 @@
...
+  # * <b>422 Unprocessable Entity</b> if the assignment is not a New Quizzes assignment
+  # * <b>503 Service Unavailable</b> if the New Quizzes service is not configured
+  #
+  # @example_response
+  #  {
+  #    "accommodations": [NewQuizAccommodation]
+  #  }
+  #
+  def index
+    return render_unauthorized_action unless authorized_to_read?
```
dbuimistrau: I've been thinking, shouldn't the `authorized_to_read?` belong to `before_action`? It looks like currently an unauthorized user can probe assignment_type and service configuration. It's not very dangerous, but since we had an incident, perhaps it would be better safe than sorry?

app/controllers/quizzes/quiz_extensions_controller.rb
```diff
@@ -145,21 +185,27 @@ def create
...
+      extension.extend_submission!
+      Canvas::LiveEvents.quiz_extension_created(extension)
+    end
 
     render json: serialize_jsonapi(quiz_extensions)
   end
 
   private
 
   def serialize_jsonapi(quiz_extensions)
```
dbuimistrau, line 155: We changed the `serialize_jsonapi` method. The problem is, it's used by both new `index` and the pre-existing `create` actions. Now the method returns raw integers instead of strings. Wouldn't it cause issues for the existing consumers?

app/controllers/quizzes/new_quiz_accommodations_controller.rb
```diff
@@ -0,0 +1,229 @@
...
+      next unless participant.is_a?(Hash)
+
+      user_id = participant["user_id"] || participant["canvas_user_id"]
+      next if user_id.nil?
+
+      extra_time = participant["extra_time"] || participant["time_extension"]
+      extra_attempts = participant["extra_attempts"]
+      reduce_choices_enabled = participant["reduce_choices_enabled"]
+
+      next unless extra_time.to_i > 0 || extra_attempts.to_i > 0 || reduce_choices_enabled == true
```
dbuimistrau: A nitpick: we use `== true` here. That means the service response of "`reduce_choices_enabled" => "true" `will be considered truthy. Normally we wouldn't care, but, as I can see, in the other places alternate key names are handled (`canvas_user_id, time_extension`). So we currently have inconsistency here.

dbuimistrau (changes requested):

app/controllers/quizzes/new_quiz_accommodations_controller.rb
jasonfill: Auth moved to before_action in NewQuizAccommodationsController — authorize_read now runs before require_assignment and require_new_quizzes_service, preventing unauthorized users from probing assignment type or service configuration.

app/controllers/quizzes/quiz_extensions_controller.rb
jasonfill: Restored original serializer for create in QuizExtensionsController. Now create uses the original serialize_jsonapi (serializer-based, stringified IDs) and index uses a new serialize_extensions method (raw hashes, integer IDs).

app/controllers/quizzes/new_quiz_accommodations_controller.rb
jasonfill: reduce_choices_enabled boolean coercion - replaced == true with Canvas::Plugin.value_to_boolean(...) to consistently handle string "true" from the NQ service, matching the defensive approach used for alternate key names (canvas_user_id, time_extension)

jasonfill:
@dbuimistrau I am closing this PR and have opened this one: https://github.com/instructure/canvas-lms/pull/2644. The changes requested are in the new PR. We needed to do this to clean up our repos master branch so we could get a few more PRs worked up and submitted off a clean master.

PR #2654, opened 2026-07-21
invite date-pending enrollments to invite_all conferences

### LINKED ISSUE #2653: Conferences with "Invite all course members" never add students when the course has a start date
### Describe the bug

Hi there,

This is Fred Dixon, the product manager for BigBlueButton.  We've had a few reports of Canvas customers hitting this bug.

I've done some digging (with the help of Fable) will submit a PR for review.

## Summary

When a conference is created with **"Invite all course members"** in a course whose start date is in the future (with **"Users can only participate in the course between these dates"** enabled), no students are ever invited.

The initial invite silently adds no students, and when the course start date passes, students are still not added. Students who enroll before the start date are also never synced, despite the automatic enrollment sync for invite-all conferences.

As a result, the instructor must edit the conference after the course begins and re-select **"Invite all course members"** before students are added.

## Additional notes

Removing the course start date makes everything work, which points to the root cause:

- `WebConference#invite_users_from_context` filters invitees through `Course#participating_users`, which uses the `enrollments.active_by_date` scope. Before the course start date, students' `enrollment_states.state` is `pending_active`, so they are filtered out when the conference is created.
- The only automatic path that adds users later is `Enrollment.after_commit :sync_to_invite_all_conferences`, which runs on `workflow_state` transitions to `active`. When a course start date passes, only `enrollment_states` changes (date-based); `enrollments.workflow_state` does not, so this callback never runs. For students enrolled before the course starts, the callback already ran at enrollment time, but the invite was filtered out by the same date-based scope.

## Possible fix (would like a review of this)

Use the `active_or_pending_by_date` scope when building conference invitees so that date-pending enrollments are invited. The existing `:join` permission already prevents users from accessing the conference before the course becomes readable.

### Screenshots or screen recordings

_No response_

### Steps to reproduce

1. As a teacher, create a course with a start date in the future and enable **"Users can only participate in the course between these dates"**.
2. Enroll one or more students and publish the course.
3. Create a conference and check **"Invite all course members"**.
4. Observe that the conference has no student invitees.
5. Enroll an additional student before the course start date.
6. Observe that the newly enrolled student is not added either.
7. Wait for (or move) the course start date to pass.
8. Observe that students are still not invitees and cannot join the conference.

## Expected behavior

Students enrolled in the course should be invited to the conference regardless of whether the course start date has passed, or at minimum should be added automatically once the course becomes active.

Access control is unaffected either way, since joining a conference already requires read permission on the course.

### What browsers are you seeing the problem on (if applicable)?

_No response_

### Backtrace

```text

```

### Relevant log output

```text

```

ffdixon:
Conferences created before a course start date silently dropped all invitees because `participating_users` filters by `enrollment_states.state = 'active'`, while students remain `pending_active` until the course start date. Since the start date transition updates only `enrollment_states` (not `enrollments.workflow_state`), conference synchronization is never re-triggered. Use the `active_or_pending_by_date` scope instead; the existing `:join` policy still prevents access until the course is readable.

fixes gh-2653
flag=none

test plan:
- create a course with a future start date and "students can only participate between these dates"
- enroll students, create a conference inviting all course members
- confirm students appear as conference invitees
- enroll another student before the start date; confirm they are added to the conference
- conclude a student; confirm they are not invited

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -4393,6 +4393,12 @@ class Course < ApplicationRecord
     User.where(id: enrollments.active_by_date.where(user_id: user_ids).select(:user_id))
   end
 
+  # like participating_users, but includes users whose enrollments are only
+  # pending because the course/term hasn't started yet (pending_active, etc.)
+  def participating_or_pending_users(user_ids)
+    User.where(id: enrollments.active_or_pending_by_date.where(user_id: user_ids).select(:user_id))
+  end
+
   def student_view_student
     User.transaction do
       fake_student = find_or_create_student_view_student
--- a/app/models/web_conference.rb
+++ b/app/models/web_conference.rb
@@ -285,7 +285,10 @@ class WebConference < ApplicationRecord
   end
 
   def invite_users_from_context(user_ids = context.user_ids)
-    members = context.is_a?(Course) ? context.participating_users(user_ids) : context.participating_users_in_context(user_ids)
+    # use a date-tolerant scope so conferences created before the course/term
+    # start date still register their invitees; the :join policy already
+    # prevents access until the course is readable by the user
+    members = context.is_a?(Course) ? context.participating_or_pending_users(user_ids) : context.participating_users_in_context(user_ids)
     new_invitees = members.to_a - invitees
     new_invitees.uniq.each do |u|
       add_invitee(u)
--- a/spec/models/enrollment_spec.rb
+++ b/spec/models/enrollment_spec.rb
@@ -4160,6 +4160,16 @@ describe Enrollment do
       end.not_to change { conference.reload.invitees.include?(user) }
     end
 
+    it "syncs enrollments that are pending because the course has not started" do
+      conference # Force lazy evaluation to create conference first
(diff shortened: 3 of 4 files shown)


spencerolson:
Hi @ffdixon thanks for this! This code has been merged and will be deployed to Production during tomorrow's scheduled deploy.

I'm sure you know this, but keep in mind that with this solution, date-pending-enrollments* still won't get conference invites if the `Restrict students from viewing course before course start date` or `Restrict students from viewing course after course end date` Course settings are checked (however, date-pending-enrollments* *will* get invites when those Course settings aren't checked). This is because the `active_or_pending_by_date` scope being used filters out "restricted access" enrollments:

```
...where(enrollment_states: { restricted_access: false })
```

if you wanted this to send invites even when those course `Restrict...` settings are checked, then you could have used the `active_or_pending_by_date_ignoring_access` scope which does not filter by `restricted_access: false`.

I think this was intentional, but just wanted to call it out!

Thanks,

Spencer

*date-pending-enrollments = enrollments in a section that has "Students can only participate in the course between these dates" checked, with a start date in the future.

